PR #13, opened 2016-09-24
git-push-pull support

namibj:
Added gitpython to requirements.

### DIFF
--- a/crowd_anki/github/github_importer.py
+++ b/crowd_anki/github/github_importer.py
@@ -34,36 +34,38 @@ class GithubImporter(object):
                                         'Path:', text='<name>/<repository>')
         if repo and ok:
             self.download_and_import(repo)
+    def download_and_import_git(self, github_repo, deck_base_name):
+        repo = None
+        try:
+            repo = Repo(Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", deck_base_name, "git"))
+            repo.remote("origin").pull()
+        except git.exc.InvalidGitRepositoryError: # New repository
+            repo = Repo.clone_from(GITHUB_REPO.format(github_repo), Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", deck_base_name, "git"))
+        AnkiJsonImporter.import_deck(self.collection, Path(repo.working_tree_dir))
 
+    def download_and_import_zip(self, github_repo, deck_base_name):
+        response = urllib2.urlopen(GITHUB_LINK.format(github_repo))
+        response_sio = StringIO.StringIO(response.read())
+        with zipfile.ZipFile(response_sio) as repo_zip:
+            repo_zip.extractall(tempfile.tempdir)
+
+        deck_directory_wb = Path(tempfile.tempdir).joinpath(deck_base_name + "-" + BRANCH_NAME)
+        deck_directory = Path(tempfile.tempdir).joinpath(deck_base_name)
+        utils.fs_remove(deck_directory)
+        deck_directory_wb.rename(deck_directory)
+        # Todo progressbar on download
+        
+        AnkiJsonImporter.import_deck(self.collection, deck_directory)
     def download_and_import(self, github_repo):
         deck_base_name = github_repo.split("/")[-1]
         try:
             try:
-                repo = None
-                try:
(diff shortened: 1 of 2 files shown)


Stvad:
I would prefer it to be in a separate branch for now. (You can change target of the PR).

crowd_anki/github/github_importer.py
```diff
@@ -34,21 +35,35 @@ def import_from_github(self):
...
-
-            deck_base_name = repo.split("/")[-1]
-            deck_directory_wb = Path(tempfile.tempdir).joinpath(deck_base_name + "-" + BRANCH_NAME)
-            deck_directory = Path(tempfile.tempdir).joinpath(deck_base_name)
-            utils.fs_remove(deck_directory)
-            deck_directory_wb.rename(deck_directory)
-            # Todo progressbar on download
-
-            AnkiJsonImporter.import_deck(self.collection, deck_directory)
+            try:
```
Stvad: You should split this one into multiple functions (e.g. one for gitpython and one for zip method)

crowd_anki/github/github_importer.py
```diff
@@ -34,21 +35,35 @@ def import_from_github(self):
...
+                repo = None
+                try:
+                    repo = Repo(Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", deck_base_name, "git"))
+                    repo.remote("origin").pull()
+                except git.exc.InvalidGitRepositoryError: # New repository
+                    repo = Repo.clone_from(GITHUB_REPO.format(github_repo), Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", deck_base_name, "git"))
+                AnkiJsonImporter.import_deck(self.collection, Path(repo.working_tree_dir))
+            
+            except git.exc.GitCommandNotFound: # Git not avaiable. Use .zip archive instead
+                QErrorMessage.qtHandler().showMessage('Error accessing the git executable.\n'
```
Stvad: At this point, as you don't provide any new functionality besides zip version and don't insure presence of git exec - I would think it would be more reasonable to silently (without prompting user, but with writing to log) fall back to zip here.

Stvad: Also anki provides it's own message helpers. (see below usage of showWarning)

crowd_anki/github/github_importer.py
```diff
@@ -34,21 +35,35 @@ def import_from_github(self):
...
-            deck_directory_wb.rename(deck_directory)
-            # Todo progressbar on download
-
-            AnkiJsonImporter.import_deck(self.collection, deck_directory)
+            try:
+                repo = None
+                try:
+                    repo = Repo(Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", deck_base_name, "git"))
+                    repo.remote("origin").pull()
+                except git.exc.InvalidGitRepositoryError: # New repository
```
Stvad: Is there a way to explicitly check for this situation?

namibj: I did not see anything regarsing explicit checking when I searched for a way to check, so is it ok do do it this way then?

crowd_anki/github/github_importer.py
namibj: Ok, I will make this silent!

namibj: I will remember for the next time I need message helpers.

crowd_anki/github/github_importer.py
namibj: I will do.

crowd_anki/github/github_importer.py
```diff
@@ -33,22 +34,38 @@ def import_from_github(self):
...
+            repo_zip.extractall(tempfile.tempdir)
 
+        deck_directory_wb = Path(tempfile.tempdir).joinpath(deck_base_name + "-" + BRANCH_NAME)
+        deck_directory = Path(tempfile.tempdir).joinpath(deck_base_name)
+        utils.fs_remove(deck_directory)
+        deck_directory_wb.rename(deck_directory)
+        # Todo progressbar on download
+        
+        AnkiJsonImporter.import_deck(self.collection, deck_directory)
+    def download_and_import(self, github_repo):
```
Stvad: space before functions. 
as a general level thing - I advise you to install PyCharm (or other IDE of your preference, but I prefer PyCharm) and follow it's style recommendations.

Issue #1, opened 2016-09-01
Port to Anki 2.1.0


nicolas-raoul:
Trying to install the plugin (1788670778) results in the following error with Anki 2.1.0a4:

```
An error occurred in an add-on.
Please post on the add-on forum:
https://anki.tenderapp.com/discussions/add-ons

Caught exception:
  File "aqt/addons.py", line 172, in accept
  File "aqt/addons.py", line 136, in install
<class 'TypeError'>: initial_value must be str or None, not bytes
```

The API has changed so plugins must be ported.
https://anki.tenderapp.com/discussions/beta-testing/180-anki-210-alpha-4

Stvad:
Hi, thank you for the feedback.
It's not API incompatibility, but python version incompatibility.
As 2.0 is stable version now, I prefer for plugin to be compatible with it. 
But it should be possible to make it cross-version compatible :) I'll look into it.

friebetill:
I have written a port for Anki 2.1.0. [friebetill/CrowdAnki](https://github.com/friebetill/CrowdAnki)
There were not many changes necessary and at first glance everything works quite well. 

I don't know if a pull request is appropriate, because the port is not compatible with Anki 2.0.

Stvad:
I haven't looked into it for a while, but I think it's totally possible to make changes compatible. 
I've briefly looked into the changes you've made and have a few comments:
* Running 2to3 conversion on the whole codebase is excessive. I wrote that code as Python3 initially and I think the only things that need to be fixed are a few imports and some unicode specific things
* It's especially true for thirdparty things - a lot of them are there because they are backports from python3 i.e. pathlib is natively present in P3 and most of the thirdparty things are consequence of using pathlib. I think the only thing beside it is uuid lib. So the only thing you need here to make things compatible is some try/except and aliasing on imports
* You removed support for AnkiHub registration for some reason
* The import fixes seems to be excessive too, the code was meant to be P2/P3 compatible and the new synthax is P3 only.

Stvad:
Thanks for looking into it though! I'd be happy to accept PR that maintains compatibility. Please see my comments above and feel free to create PR to discuss things in more details codewise.

evolverine:
Thanks for working on this, guys. Will be very useful when it's out. Good luck!

avidrucker:
Is this closable now?

PR #26, opened 2018-03-15
Port to Anki 2.1.0

friebetill:
Adds compatibility for Anki 2.1.X and maintains compatibility with Anki 2.0.X. #1 

Thank you for your feedback @Stvad .
I tried to work it in as best I could.

- Instead of using 2to3, I rewrote the code base by hand. I changed the imports from absolute to relative, because Python 3 is no longer compatible to absolute paths. 
- I took AnkiHub Support back in, thought it would not be compatible to Python 3, but it's in a try-catch block anyway.

I hope you're comfortable with the changes, if not, I'll be happy to make more adjustments.

### DIFF
--- a/crowd_anki/anki_exporter.py
+++ b/crowd_anki/anki_exporter.py
@@ -4,10 +4,10 @@ import shutil
 
 import anki.utils
 
-from thirdparty.pathlib import Path
+from .thirdparty.pathlib import Path
 
-from crowd_anki.utils.constants import DECK_FILE_EXTENSION, MEDIA_SUBDIRECTORY_NAME
-from crowd_anki.representation.deck import Deck
+from .utils.constants import DECK_FILE_EXTENSION, MEDIA_SUBDIRECTORY_NAME
+from .representation.deck import Deck
 
 
 class AnkiJsonExporter(object):
--- a/crowd_anki/anki_exporter.py
+++ b/crowd_anki/anki_exporter.py
@@ -42,8 +42,7 @@ class AnkiJsonExporter(object):
                                        default=Deck.default_json,
                                        sort_keys=True,
                                        indent=4,
-                                       ensure_ascii=False,
-                                       encoding="utf8"))
+                                       ensure_ascii=False))
 
         self._save_changes()
 
--- a/crowd_anki/anki_exporter.py
+++ b/crowd_anki/anki_exporter.py
@@ -4,7 +4,7 @@ import shutil
 
 import anki.utils
 
-from .thirdparty.pathlib import Path
+from .utils.pathlib_wrapper import Path
 
 from .utils.constants import DECK_FILE_EXTENSION, MEDIA_SUBDIRECTORY_NAME
 from .representation.deck import Deck
--- a/crowd_anki/anki_exporter_wrapper.py
+++ b/crowd_anki/anki_exporter_wrapper.py
@@ -1,4 +1,4 @@
-from .thirdparty.pathlib import Path
+from .utils.pathlib_wrapper import Path
 
 import anki.exporting
 import aqt.utils
--- a/crowd_anki/anki_exporter_wrapper.py
+++ b/crowd_anki/anki_exporter_wrapper.py
@@ -1,4 +1,3 @@
-#from . import crowd_anki # Unnecessary?
 from .thirdparty.pathlib import Path
 
 import anki.exporting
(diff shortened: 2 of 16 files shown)


Stvad:
Thank you, I'll look it over on the weekends!

crowd_anki/anki_exporter.py
```diff
@@ -4,10 +4,10 @@
 
 import anki.utils
 
-from thirdparty.pathlib import Path
+from .thirdparty.pathlib import Path
```
Stvad: You forgot to do aliasing here. Also instead of doing it everywhere - you can do it in one place (i.e. thirdparty/pathlib.py) and then import it from there

crowd_anki/utils/utils.py
```diff
@@ -53,6 +56,6 @@ def fs_remove(path):
         return
 
     if path.is_dir():
-        shutil.rmtree(unicode(str(path)))
+        shutil.rmtree(str(path))
```
Stvad: you did unicode aliasing but forgot to put the call to it back

crowd_anki/main.py
```diff
@@ -1,8 +1,11 @@
-from thirdparty.pathlib import Path
-
-from crowd_anki import anki_exporter_wrapper  # Do not remove. To hook exporters list extension
-from crowd_anki.anki_importer import AnkiJsonImporter
-from crowd_anki.github.github_importer import GithubImporter
+try:
+    from pathlib import Path
+except ImportError:
+    from .thirdparty.pathlib import Path
```
Stvad: see comment above about putting aliasing in one place

crowd_anki/anki_importer.py
```diff
@@ -1,13 +1,17 @@
 import json
 import shutil
 import os
-from thirdparty.pathlib import Path
+from .thirdparty.pathlib import Path
```
Stvad: missing aliasing?

crowd_anki/anki_exporter_wrapper.py
```diff
@@ -1,12 +1,12 @@
-import crowd_anki
-from thirdparty.pathlib import Path
+#from . import crowd_anki # Unnecessary?
```
Stvad: well remove it then 😛

crowd_anki/anki_exporter.py
```diff
@@ -42,8 +42,7 @@ def export_deck_to_directory(self, deck_name, output_dir=Path("."), copy_media=T
                                        default=Deck.default_json,
                                        sort_keys=True,
                                        indent=4,
-                                       ensure_ascii=False,
```
Stvad: does this still work properly on 2.0 if your system default encoding is not utf8?

friebetill: I think so, because the default parameter is set to UTF-8. 
Source: https://docs.python.org/2/library/json.html

crowd_anki/anki_exporter.py
friebetill: Thanks, I tried that, the problem was that it covered the default library with pathlib.py. 
This is why I created a pathlib_wrapper in the utils folder.

crowd_anki/utils/pathlib_wrapper.py
```diff
@@ -0,0 +1,4 @@
+try:
+    from pathlib import Path
+except ImportError:
+    from ..thirdparty.pathlib import Path
```
Stvad: thirdparty.pathlib is itself a wrapper, I'm not sure I understood why you can't reuse it, but if you can't, you might as well delete it and do aliasing that it performs directly here

Stvad:
Thank you again. I'll package both versions and play with them sometime next week and then will merge this and put out new release/upload them on ankiweb.

Stvad:
Testing it now:
* The __init__.py you've created should be within crowd_aniki directory for it to work properly in 2.1 afaik (as only that dir and crowd_anki_importer.py get packaged).

Other then that it seems to be working fine, at least on basic use cases.

friebetill:
Thanks for your quick answers and tips. 
Shall I add your suggestions or have you already done it yourself?

I would also like to thank you again for this great addon.

Stvad:
Haven't done that yet, (besides quick hack for 2.1 ), so go ahead

Stvad:
I published the updated version of the plugin on AnkiWeb

PR #54, opened 2019-04-15
Capitalise names of products & formats

katrinleinweber:
See diff ;-) I didn't update occurrences in code objects, just in user-facing docu & messages.


Stvad:
My excuse is that Capital letters are hard 😛. Thank you! 🙂

Issue #15, opened 2016-12-28
Deck name vs. repository name


axelboc:
- On export, CrowdAnki outputs a folder named after the deck - e.g. _My Deck_ becomes `My_Deck`.
- To import from GitHub, CrowdAnki requires the repository to be named after the deck - e.g. `My_Deck`.

I'm not very keen on having the name of my GitHub repo bound to the name of the deck. I think the add-on would really benefit from a bit more flexibility in this regard.

I can think of a couple of options:

1. Include the name of the deck in the JSON and allow exporting the deck to any folder (i.e. the export doesn't create the folder).
    - **Upside:** nice separation of concern; during the import CrowdAnki can just look for the JSON file and the `media` folder at the root of the repo and parse the content of the file to get the name of the deck.
    - **Downside:** changing the format is not ideal... although for backward compatibility, if the JSON file doesn't contain the name of the deck, its filename could be used as fallback.
2. Keep the export as is, but allow selecting a custom folder inside the git repository during the import.
    - **Upside**: probably more straightforward to implement; no change to the JSON format.
    - **Downside**: superfluous folder nesting; makes import more complex for end users.

Stvad:
Hi, I'm not sure I'm following what is your concern.
The name of the deck is already included in the JSON. And correspondingly code, internally, can handle any correct JSON file (though admittedly there is no UI to do that).
I've chosen to enforce the convention for deck directories to be the representation of the deck name as it was easier to implement and I haven't seen any particular advantage in the 'pick first json file strategy' (though it can be changed without breaking thins).
I've chosen to enforce convention name of the repository==the name of the deck directory for the same reason (and also to reinforce the notion one-repository=one-deck). I can see how the name of the repository may not be the nicest one as a consequence, with hierarchy and platform-specific name limitations.
Do I understand correctly that you would prefer mentioned 'pick first json file strategy' to be in place (that would imply that the repositories can have any valid name as long as they have the correct content in place)?

axelboc:
Ah sorry, I can indeed see the name of the deck in the JSON - not sure how I missed that. It makes my second option irrelevant...

So what I'm suggesting is to remove the first convention you mention that the deck directory has to match the name of the deck.

I think the import can still work the same way: select a folder and look for a JSON file with the same name. What that name is doesn't matter. No need for a 'pick first json file strategy'.

The problem is with the export. A solution might be to have a checkbox that says something like "Create deck directory": 

[...]

PR #56, opened 2019-04-19
#15 json naming convention

evolverine:
I'm suggesting a naming convention, where all the json files are named 'deck.json', which can simplify naming issues especially for the importer. Then it's both the folder they're in and also the github repo that offer the user the knowledge of which deck they have there.


Stvad:
This, makes sense, but I'd like to ensure backward compatibility on import. I.e. the import with the old convention should still work.

crowd_anki/export/anki_exporter.py
```diff
@@ -28,7 +28,7 @@ def export_to_directory(self, deck: AnkiDeck, output_dir=Path("."), copy_media=T
         deck = deck_initializer.from_collection(self.collection, deck.name)
         self.last_exported_count = deck.get_note_count()
 
-        deck_filename = deck_directory.joinpath(deck_fsname).with_suffix(DECK_FILE_EXTENSION)
+        deck_filename = deck_directory.joinpath('deck').with_suffix(DECK_FILE_EXTENSION)
```
Stvad: Please extract the actual name used to a constant in https://github.com/Stvad/CrowdAnki/blob/master/crowd_anki/utils/constants.py

crowd_anki/importer/anki_importer.py
```diff
@@ -42,7 +42,7 @@ def load_from_directory(self, directory_path, import_media=True):
             aqt.mw.progress.start(immediate=True)
 
         try:
-            self.load_from_file(directory_path.joinpath(directory_path.name).with_suffix(DECK_FILE_EXTENSION))
+            self.load_from_file(directory_path.joinpath('deck').with_suffix(DECK_FILE_EXTENSION))
```
Stvad: You should update the docstring of the function to describe new convention

Stvad:
Btw, I see several merge commits on the branch - please look into using the `git pull --rebase` instead to maintain the git history cleaner

Stvad:
Also, the Readme should be updated to reflect new convention

evolverine:
Apologies for the delay. I will look at implementing your suggestions next weekend.

evolverine:
> 
> 
> This, makes sense, but I'd like to ensure backward compatibility on import. I.e. the import with the old convention should still work.

I've implemented this in the latest commit (and tested it too, locally; seems to work just fine).

crowd_anki/export/anki_exporter.py
evolverine: good point, done in the latest commit.

crowd_anki/importer/anki_importer.py
evolverine: done in the latest commit.

evolverine:
> 
> 
> Also, the Readme should be updated to reflect new convention

I've checked the Readme, and the only places that the file naming convention comes up seems to be in the screenshots. And I haven't changed them, partly because I didn't want my git username and projects to replace yours; it only makes sense that they reflect your workspace, given that you are the author of the plugin, at least that's what I reasoned. But let me know if you'd like these updated too, and I can do that.

Stvad:
Hey, sorry for the delay in looking at this. I've checked it and merged it with minor updates

evolverine:
Great, thank you!

Issue #23, opened 2017-10-31
Make ‘respecting the movement of cards between decks’ optional


ELLIOTTCABLE:
The documentation states that, if a card is in a different deck when a CrowdAnki deck is imported, that it will be moved in Anki as well.

This should be optional: decks should be able to be, optionally, ’virtual‘ — that is, a “Foo” deck with 24 cards is imported, but then those cards are *moved* by the user into another “Major” deck with hundreds of other cards, then re-importing “Foo” should simply update those notes/cards where they already exist: in the “Major” deck.

The issue here is scalability: Anki (as mentioned in the manual) doesn't handle large numbers of decks, well. It's pretty standard practice to distribute a *subset* of information in a “shared deck”, but usually, one *merges* many of those into a single deck *per subject* (or writes their own cards about that subject, that don't belong in the original shared deck, or … you get the idea.) Thus, it makes sense for (most) CrowdAnki decks to be ‘virtual’ like that: import them, merge them into your *own* subject-specific decks, and then merge upstream changes as they're offered.

(Really the only situation where one wants ‘concrete’ decks, ones that precisely match the CrowdAnki representation, and follow said representation when it's imported if it's changed, is when one is using CrowdAnki to backup/record one's entire collection — and that's currently not well-supported in other ways, so …)

Just some thoughts! `(=`

Stvad:
Maybe I'm missing something, but why do you need to merge the decks if you can have a hierarchy of them?

ELLIOTTCABLE:
@Stvad cf. <https://apps.ankiweb.net/docs/manual.html#using-decks-appropriately>

ohare93:
People do not use CrowdAnki to merge _decks_ but to merge _notes_. Using this type of system allows us to manage and maintain the **content** of the notes, owever which deck each user decides to put those notes

Stvad:
Also discussed more in #8 with a fork there proposing a prototype for basic way to make this configurable (https://github.com/dayjaby/CrowdAnki/commit/39c9666d58b16483631c13521597ae38a9d66cfa)

ohare93:
I have submitted PR (#69) which fixes this issue 👍 @ELLIOTTCABLE

ohare93:
(pinging @dayjaby too, as he complained about this in Issue #8)

ohare93:
Issue is now fixed 👍 there is now an option in the config settings to not move notes on import (see PR #69)

PR #69, opened 2019-11-11
Ignore Deck Movement Config Settings Option

ohare93:
Can't stop won't stop :musical_note: Fix for Issue #23 :+1: 

Had to pass config around in a lot of places :sweat_smile: it may have just been easier to call the static version where I needed it, but I imagine this work will be used later on as well for other such config options.

All tests passing, have manually tested import/export without the feature, and the feature itself works for me :grin:


crowd_anki/representation/note.py
```diff
@@ -117,5 +123,5 @@ def save_to_collection(self, collection, deck, model_map_cache):
 
         if new_note:
             collection.addNote(self.anki_object)
-        else:
+        elif not self.config.import_notes_ignore_deck_movement:
```
Stvad: there is too much of passing things around, I'd prefer if you use a pattern similar to what we've done with `addonManager` here. (making the config one of the `__init__` parameters with None as default value and initializing it via singleton value if nothing is passed in)

crowd_anki/config/config_dialog.py
```diff
@@ -32,6 +32,7 @@ def accept(self):
             super().accept()
 
     def ui_initial_setup(self):
+        # Snapshot Options
```
Stvad: better just extract each of those sections into a separate method

ui_files/config.ui
```diff
@@ -58,6 +58,22 @@
...
+       <item>
+        <widget class="QGroupBox" name="group_deck_import">
+         <property name="title">
+          <string>Import</string>
+         </property>
+         <layout class="QVBoxLayout" name="verticalLayout_5">
+          <item>
+           <widget class="QCheckBox" name="cb_ignore_move_cards">
+            <property name="text">
+             <string>Do Not Move Existing Cards</string>
```
Stvad: I think this warrant a more detailed explanation, potentially with a reference to a github issue

crowd_anki/representation/note.py
ohare93: Yea thought you may say that, no worries. I am inclined to agree :sweat_smile: Done :+1:

crowd_anki/config/config_dialog.py
ohare93: Done :+1:

ui_files/config.ui
ohare93: Done :+1: give it a look

crowd_anki/config/config_dialog.py
```diff
@@ -32,6 +32,11 @@ def accept(self):
             super().accept()
 
     def ui_initial_setup(self):
+        self.ui_setup_snapshot_options()
```
Stvad: I don't think we need the `ui` in the name, we're in the ui widget after all ;).

crowd_anki/representation/json_serializable.py
```diff
@@ -115,7 +115,7 @@ def from_json(cls, json_dict):
 
 
 class JsonSerializableAnkiObject(JsonSerializable):
-    export_filter_set = JsonSerializable.export_filter_set | {"anki_object", "anki_object_dict"}
```
Stvad: this should be in the `Note` filter

crowd_anki/representation/note.py
```diff
@@ -38,8 +41,8 @@ def from_collection(cls, collection, note_id, note_models):
         return note
 
     @classmethod
-    def from_json(cls, json_dict):
-        note = Note()
+    def from_json(cls, json_dict, config: ConfigSettings = None):
+        note = Note(config=config)
```
Stvad: why do we need these?

crowd_anki/representation/json_serializable.py
ohare93: Okay, done :+1:

crowd_anki/representation/note.py
ohare93: Well I added an optional config parameter to the Note constructor, and these are the only functions which ever call the Note constructor. They are used throughout the code, as the default way to make notes, so I thought they also needed the ability to have this config item passed to them :thinking: No one is ever going to simple call Note()

crowd_anki/config/config_dialog.py
ohare93: Yea sure why not :ok:

crowd_anki/representation/note.py
Stvad: eh, this argument have some merit, but I'm more of a mind that the code that is not used should not be there. Do you envision this being used any time soon?

ohare93: Nope, not in the slightest! I shall remove it then :wink:

crowd_anki/representation/note.py
```diff
@@ -28,7 +32,7 @@ def get_notes_from_collection(collection, deck_id, note_models):
     @classmethod
     def from_collection(cls, collection, note_id, note_models):
         anki_note = AnkiNote(collection, id=note_id)
-        note = Note(anki_note)
+        note = Note(anki_note=anki_note)
```
Stvad: 👀

ohare93: :thinking: ?

The Note constructor now has two optional parameters:

```python
    def __init__(self, anki_note=None, config: ConfigSettings = None):
        super(Note, self).__init__(anki_note)
        self.note_model_uuid = None
        self.config = config or ConfigSettings.get_instance()
```

So I made sure to be explicit about which optional parameter:
```python
    @classmethod
    def from_collection(cls, collection, note_id, note_models):
        anki_note = AnkiNote(collection, id=note_id)
        note = Note(anki_note=anki_note)
```

Stvad: they are still positional though 🤷‍♂️

ohare93: _Readable and explicit!_ 😁 🤷‍♂

ohare93:
Pulled already? That was easy 😁 👍

Stvad:
yep, thank you for you contributions ;)

Issue #62, opened 2019-08-11
Feature Request - Ignore Field on Import/Export


ohare93:
I love the collaboration that this add-on allows, but there is one thing that is stopping me from being able to share my decks with others. I want to begin sharing some of my personal language learning decks with others, however **I use images in my cards**. Using images is wonderful for memory, there's nothing better. But when my deck is filled to the brim with copyrighted images simply taken from Google Images for personal use, this makes my decks completely unshareable. Of course it would be better if I had copyright free images (which I do for some), but finding that for thousands of notes, especially vague concepts, is incredibly challenging.

_Potential Solution:_ The ability to flag any field(s) so that _their value_ will not be exported/imported. This would allow each user to set their own picture, while still being able to import the deck when it is upgraded _without their manually set "Picture" field being overwritten_. Exporting without certain fields would be nice, but I will be using other deck managers like AnkiDM, so Importing really is the main issue.

I imagine this could be implemented for importing in many ways. 

1) State that the field should not import in the Note Model. The field should still be taken into account so you can match it to the correct Note Type, but it's value will not be put into the Note itself (thus leaving anything the user has set changed). 

```
  "note_models": [
    {
      "__type__": "NoteModel",
      "crowdanki_uuid": "43e2586a-9a65-11e8-a777-a0481cc15658",
      "name": "Ultimate Geography",
      "flds": [
        {
        ...
        {
          "name": "Map",
          "ord": 7,
          "font": "Arial",
          "media": [],
          "rtl": false,
          "size": 20,
          "sticky": false
          "doNotImport": true        <-----------------------------
        }
      ],
```

2) State that the specific card does not overwrite specific fields, with some type of generic flag which CrowdAnki reads on Import.

```
  "notes": [
    {
      "__type__": "Note",
      "data": "",
      "fields": [
        "England",
        "Constituent country of the United Kingdom.",
        "London",
        "",
        "Constituent country",
        "<img src=\"ug-flag-england.svg\" />",
        "",
        false                 <----------------------------------
      ],
      "flags": 0,
      "guid": "e+/O]%*qfk",
      "note_model_uuid": "43e2586a-9a65-11e8-a777-a0481cc15658",
      "tags": [
        "UG::Europe"
      ]
    },
```

[...]

Issue #31, opened 2018-08-07
Export/import a deck without configuration


axelboc:
1. Let's say I have a deck in Anki with configuration **A** = 10 new cards + 100 reviews / day.
2. I export the deck with CrowdAnki.
3. I switch the deck to configuration **B** = 5 new cards + 50 reviews / day.
4. I re-import the deck.
5. **Deck has configuration A again instead of B**.

This behaviour is problematic notably when users try to update a deck to a newer version. Is there any way exporting/importing a deck could _not_ export/import the deck configuration? I tried removing keys `deck_configurations` and `deck_config_uiid` from the exported JSON before re-importing it, but I get an error 😢 :

```
Traceback (most recent call last):
  File "...\addons\crowd_anki\main.py", line 23, in on_import_action
    AnkiJsonImporter.import_deck(aqt.mw.col, import_directory)
  File "...\addons\crowd_anki\anki_importer.py", line 68, in import_deck
    importer.load_from_directory(directory_path, import_media)
  File "...\addons\crowd_anki\anki_importer.py", line 44, in load_from_directory
    self.load_from_file(directory_path.joinpath(directory_path.name).with_suffix(DECK_FILE_EXTENSION))
  File "...\addons\crowd_anki\anki_importer.py", line 27, in load_from_file
    deck = Deck.from_json(deck_json)
  File "...\addons\crowd_anki\representation\deck.py", line 154, in from_json
    deck.deck_config_uuid = json_dict["deck_config_uuid"]
KeyError: 'deck_config_uuid'
```

Stvad:
I don't think you can get there by modifying JSON. To get this functionality, this has to be made explicitly configurable

johnlawrenceaspden:
It does seem really bad to fiddle with the configurations, they're not part of the deck, they're part of how the particular anki user is using the deck. Two people sharing the same deck might want different learning rates, e.g.

It would probably be best to drop this config/import export entirely, and if you import a new deck, give it the user's default configuration.

Stvad:
I see that this can be problematic, but I also see the value in the "full export", so the deck would be fully self-contained.  
I believe that making this configurable would be valuable

gabbard:
@Stvad : I second @johnlawrenceaspden that this can be a big problem; in my case I'm trying to coordinate an evolving deck between five people, all of whom need different deck configurations because they are different ages!

I would be happy to take a stab at making this configurable when I have some free time in mid-November, if no one else has fixed it by then.

[...]

PR #72, opened 2019-11-18
Personal Fields + Import Config Window

ohare93:
Hello again! :grin: I come to you today with a change which is very important to me: the ability to be able to import a CrowdAnki json file with some fields set as "Personal" which then keep the old field data before the import. Sort of like a merge :thinking: this would fix my issue #62 

However this is just a small proof of concept, only 1 file is changed! :sweat_smile: while brewing over how to do this change I came to the conclusion that this is a very fiddly issue which could be done in many different ways. Therefore I'd rather come to you early in the process and lay out the options and my thoughts. Perhaps there's an easier way I was not seeing :smile: :smirk: Regardless, I'll save myself some hassle down the line by doing it this way now, I feel.

---

### Proof of concept
It is rather simple, as each field in an Anki flashcard is serialized to a string in CrowdAnki json, even if it is blank. I simply added a check for if a field value is false, and if so I overwrite it with the current value of that field before updating the card as a whole :+1: This is a rather simplistic approach, and perhaps may be considered hack-y, but it gets the job done.

So the current implementation is done *on a per note basis* and currently only works on importing (with a field set to false). This has many downsides, and I would prefer the ability to set an entire field as "Personal", as that would allow for: 
- Export notes without "Personal" fields
- Default values on first import of a card

The first one can simply be an export option to select, but for a deck with many many types of cards it can get annoying. Not to mention if a hosted deck has some "Personal" fields, it should be simple and seamless to simply export those as "Personal" too, as the default.

So how to set a whole field as "Personal" is the tough question :thinking: :face_with_head_bandage: Here are some ideas:
1) Check for a special string start/end of the field name, such as "PERSONAL_Picture" or "NotesPERSONAL".
Good for hosted decks, bad for user choice and clarity.
1) Get each user to set the "Personal" fields in the Config settings.
Good for user choice, bad that they have to set it up for each deck they download.
1) A mixture of both: a flag of "Personal: true" on the deckmodel, which the importer checks and registers this notemodel as having that field as "Personal" in each user's config :thinking: 

For exporting there can always be a tickbox "Override Personal Fields", or some such, which allows you to set more or less fields as "Personal" than what is in your config.

Anyways, that's my brain dump over :sweat_smile: Let me know what you think, whenever you have time :+1:


Stvad:
As a general comment - look into using rebase instead of merges for git to keep the history cleaner

Stvad:
I think it'd be a good idea to outline the use-cases here, which would help with design decisions

**Things I've heard so far:**
* Personal notes field
* Copyrighted material field (This one is confusing for me, as presumably you'd want the exclusion here to work not on the field level but on individual notes level, as you still want to include the things with an appropriate license?)

**Not entirely sure if possible, but may be neat:**
* `Local override field` - example would be a deck that has shortcuts for some program, and you may want to customize shortcuts. Then you'd add your custom shortcuts in the override field. Note display logic would need to have a way to conditionally display things on cards though.

Stvad:
Given that fields are defined in the NoteModel - it can be a better vessel for the field type information (so my default intuition would be closer to the option 1 you've outlined on #62). + Allowing user choice on how to deal with personal fields (i.e. have import/export personal fields config/checkbox on corresponding dialogs)

There is a question of transferring that information between Anki instances [this discussion](https://anki.tenderapp.com/discussions/add-ons/19030-how-does-anki-handle-custom-data-in-collection) may be relevant

Stvad:
Regarding the poc - It works 🎉. I'm not sure if doing things on per note basis scales though ;) Also I'd probably prefer to have the explicit configuration over convention in this case

ohare93:
> As a general comment - look into using rebase instead of merges for git to keep the history cleaner

Yea sorry, will do 😭

ohare93:
> I think it'd be a good idea to outline the use-cases here, which would help with design decisions

Definitely! 👍 

> Things I've heard so far:
> 
> Personal notes field
> Copyrighted material field (This one is confusing for me, as presumably you'd want the exclusion here to work not on the field level but on individual notes level, as you still want to include the things with an appropriate license?)

I think it would be better for it to be on the field level, to just have a {{Picture}} field which does not sync to the users, but that they can set themselves. I am not against the concept of a doing it on a per note basis, as that may be useful in some other way, but I'd like it to be settable in the notemodels 👍

ohare93:
Since I have a POC version of this I've been busy with other stuff that will actually use this. But just to be clear, I am open to doing it whatever way 🤷‍♂ your preference would be for it to be on the notemodel, and I am fine with implementing that. 

However my main goal is to get *importing* personal fields functional. It would also be nice to be able to export, and to have saved export settings, and to offer to automagically set those export settings when you import a deck with personal fields, etc, etc. However all of that comes with extra issues and complexity, and is not *needed* for me right now 😅 

So I suppose I am asking: if I implement support of a Personal Field just in the importing of a CrowdAnki json file (by adding an entry to the corresponding Notemodel definition), is that good enough as an MVP of this feature? I am happy to extend it to Exporting, then be smart and save user settings, etc, but not in the next few months, I expect 😅

Stvad:
You can definitely work on things incrementally. Start by implementing importing on NoteModel level, and it can be marked as experimental feature initially. And then work towards full end-end support over time :)

ohare93:
Work in progress, was delayed greatly due to my Anki development build being broken. Spent two days on it, but have just given up to actually get something useful done. Who even needs breakpoints? :sob: 

Anyways, I have created an Import window, and interrupted the flow of Import to display it before the actual importing begins. The current idea of the implementation is that it reads in the "_settings" key from the json (open to a different name! :sweat_smile:) then displays that to the user for them to update in a proper UI window, then after the user accepts their import settings it does the import as normal, passing those new values to the importer to use in various places. 

I also have a sense that the exporter will be the exact same in terms of settings, so I plan to just join them together into one import/export settings window. On export it reads in the settings from the existing config (such as sort method, etc) and allows the user to do a one time change before export. On import it does the same, but also reads in the settings from the json file itself and performs any overrides.

Thoughts behind this structure as a plan? :thinking: 

P.S: This import window is quite powerful as a concept, I plan to add a few extra goodies into there after Personal Fields :wink:

Stvad:
@ohare93 Welcome back! :)  I haven't done it in a while, but I remember setting up the debugging being challenging =\

A few thoughts from your description:
* After the import is complete, where/how do we store the value of `settings` (as it's not imported to the anki db)?
* I don't want to see the dialog each time I do import/export. I think there should be a way to say "accept whatever is in settings by default"  (a meta-setting?)

ohare93:
> * After the import is complete, where/how do we store the value of `settings` (as it's not imported to the anki db)?

The values are not intended to be saved, just passed down through the import/export logic to change the data in various ways. It essentially says "here are the import/export setting values stored in **both your config and the import file itself**, do you want to continue with these settings?"

> * I don't want to see the dialog each time I do import/export. I think there should be a way to say "accept whatever is in settings by default"  (a meta-setting?)

Well it's just one popup with nothing you *have to do* on it, so you can just click ok if you really don't care. Having that show up once on import is not a big deal in my view, and a "don't ask me about it again setting" kind of breaks the concept of "here are the values stored in the import file" :disappointed_relieved: 

Some of the settings I would like for deck managers to be able to set include:
* Personal Fields (already discussed :wink:) presets, which the user can disable individually, or even add more before import
* Inform user of changes to notes or note structure (see https://github.com/axelboc/anki-ultimate-geography/issues/306#issuecomment-623484662). Notes may have been deleted since their last update, or Note Models have changed. A place for a small explanatory message would go a long way.
    * Offer to tag all the imported notes, so the user can find the ones they may want to delete.
    * Tickbox for "Duplicate Note Model" in case they want to keep the old version, or want all their old notes to say on the old version. (Technically done now when there is a detected change, as the old note models aren't deleted, right? :thinking: having it as an option to do at any point would be useful, regardless)

And some options it would be good to give the user in this window on import/export:
* Import/Export Headers?
* Import/Export Headers?

ohare93:
Update: Working PR :+1: 

The new Import Window looks like this!

image

The general idea is that this data is prefilled with data from the import file (and the user's default settings in the config), then the user has the option to change each of them for this individual import. On pressing Okay the app continues the normal import flow, but a new variable `import_config` (open to a new name :sweat_smile:) is passed down to each of the possible affected parts.

# Personal Fields

The Personal Fields section of the window is autofilled with all the note models used in the import, showing all their fields. The fields are ticked by default if it finds a `"personal_field": true` inside the note_model field of the import.

```
                {
                    "font": "Liberation Sans",
                    "media": [],
                    "name": "Picture",
                    "ord": 2,
                    "personal_field": true,     <---------------------------
                    "rtl": false,
                    "size": 6,
                    "sticky": false
                },
```

Users can then deselect or select each of the fields and choose import. Each field that is selected as a Personal Field is added into a list of Tuples (note_model, field_name), and each card checks if it's own fields are in the list on save to collection (see note.py/handle_dictionary_update for the specifics).


# Everything Else

I added a few extra features into this window as it was quite bare, and I wanted to actually settle on a good setup for passing an `import_config` with multiple values, rather than just one `personal_fields` object. Making these other values work was a piece of cake. **I also hope these show the power of this import structure!** :grin: 

## Import Message

The import message is just a way for a deck manager to communicate with a deck user on import, to warn of potential changes, inform what they are, or whatnot. This can be set in the `_settings` value of a CrowdAnki import, but will function perfectly fine without it.

```
    "_settings": {
        "import_message": "This release has deleted cards X, Y, and Z. Please sort that out for yourself using blahblahblah.",
        "suggest_tag_imported_cards": true
    }
```

## Tag Imported Cards

Tick this box and fill in a comma separated list of tags. Each card imported will receive that tag, as well as the ones on the cards themselves. The hope is that users will be able to more easily be able to separate cards imported in this run from others not affected, or perhaps deleted cards. 

There is also a "suggest_tag_imported_cards" option that can be set in the `_settings` which came about as an idea from the discussion here: https://github.com/axelboc/anki-ultimate-geography/issues/306#issuecomment-623484662 All this does is auto tick this box and add "Suggested by deck manager!" to the title. 

## Do Not Move Existing Cards

The config value I added a while ago, but now you can set it for each import. Defaults to the value in the ConfigSettings.

## Deck Parts to Use

**Not yet functioning!** Apart from media, that works. The hope is that you can simply import the parts of the CrowdAnki import that you want, ignoring the rest. But this one may need some gooood testing and configuration :sweat_smile: just left it here to show it as a possibility, and am happy to disable the non working ones should we get to PR stage soon.



Anyways that's all for now, I am happy to have a working and usable version. Also what do people think about PersonalFields being inside the NoteModel? :thinking: I did that at the start, thinking I had to put it there, but it could just as easily go inside the new `_settings` value as a dictionary. Thoughts?

crowd_anki/representation/json_serializable.py
```diff
@@ -6,8 +6,8 @@
 
 class JsonSerializable:
     readable_names = {}
-    export_filter_set = {"mod",  # Modification time
-                         "usn",  # Todo clarify
+    export_filter_set = {  # "mod",  # Modification time
```
Stvad, line 9: don't forget to revert hotfix changes :)

ohare93: Of course :wink:

ui_files/import.ui
```diff
@@ -0,0 +1,194 @@
...
+           <widget class="QCheckBox" name="cb_ignore_move_cards">
+            <property name="text">
+             <string>Do Not Move Existing Cards</string>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <widget class="QLabel" name="label">
+            <property name="text">
+             <string>Personal Fields - [Explanation of what that is]</string>
```
Stvad, line 133: um, I imagine this should actually contain explanation? 😛

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,126 @@
...
+                field_ui.setCheckState(Qt.Checked if is_field_personal(field) else Qt.Unchecked)
+                field_ui.setFlags(Qt.ItemIsEnabled | Qt.ItemIsUserCheckable)
+                self.form.list_personal_fields.addItem(field_ui)
+
+    def setup_misc(self):
+        self.form.import_message_textbox.setText(self.import_settings.import_message)
+
+        if self.import_settings.suggest_tag_imported_cards:
+            self.form.cb_tag_cards.setCheckState(Qt.Checked)
+            self.form.cb_tag_cards.setText("Tag Cards (Suggested by Deck Manager!)")
```
Stvad, line 96: what's deck manager? (I may guess that you're talking about `anki-dm` or some-such, but I don't want that to be a prerequisite knowledge for using CrowdAnki)

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,126 @@
...
+        for model in self.deck_json["note_models"]:
+            heading_ui = QListWidgetItem(model["name"])
+            heading_ui.setFlags(Qt.ItemIsEnabled)
+            self.form.list_personal_fields.addItem(heading_ui)
+
+            for field in model["flds"]:
+                field_ui = QListWidgetItem(field["name"])
+                field_ui.setCheckState(Qt.Checked if is_field_personal(field) else Qt.Unchecked)
+                field_ui.setFlags(Qt.ItemIsEnabled | Qt.ItemIsUserCheckable)
+                self.form.list_personal_fields.addItem(field_ui)
```
Stvad, line 89: should be separate function

ui_files/import.ui
```diff
@@ -0,0 +1,194 @@
...
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout">
+     <item>
+      <layout class="QVBoxLayout" name="verticalLayout_2">
+       <item>
+        <widget class="QGroupBox" name="group_deck_import">
+         <property name="title">
+          <string>Import</string>
+         </property>
```
Stvad, line 25: this label at the top feels redundant

ui_files/import.ui
```diff
@@ -0,0 +1,194 @@
...
+             <number>6</number>
+            </property>
+            <item>
+             <layout class="QHBoxLayout" name="horizontalLayout_2">
+              <item>
+               <layout class="QVBoxLayout" name="verticalLayout_4">
+                <item>
+                 <widget class="QLabel" name="import_message_label">
+                  <property name="text">
+                   <string>Import Message</string>
```
Stvad, line 39: should be more emphasized (bold?)

Stvad: true for other panel headers too

crowd_anki/representation/json_serializable.py
ohare93: Done.

ui_files/import.ui
ohare93: Holdover from previous work. Removed.

ui_files/import.ui
ohare93: Bolded headers :+1:

ohare93:
> * On reflection I think it'd be preferable to separate the meta information into distinct section of the JSON file (or maybe even a separate file, it can be Yaml then :p) instead of mixing it in the field directly.  So you can have something like

Sounds like a great idea :+1: Will be must easier for people to manually edit. I can whip that up nice and quick :grin: 

> Added some comments on the XML for this. All headers could use emphasis.

Done.

> Also the distinction between different note models should be more distinct in the list (headers emphasized, adding separators. Also should be clearer that different note models is what we're seeing

Yea, it could be better. I'll see what I can do here :+1:

ohare93:
Alright, the import_config is now a Yaml file :+1: and is fully working again. I also refactored the setup of Personal Fields, to be more functional.

The import file I am using is:
```

import_message: This release has deleted cards X, Y, and Z. Please sort that out for yourself using blahblahblah.
suggest_tag_imported_cards: yes


note_models:
    LL Noun:
        personal_fields:
            - Picture
            - Plural
    LL Verb:
        personal_fields:
            - Picture
```

You will notice that I went with the model name, and not the uuid :thinking: I think it is better to use the name of the model, as it will more easily enable users to read and understand the file they are importing (both importers and deck managers). What do you think on this? :slightly_smiling_face:

crowd_anki/utils/constants.py
```diff
@@ -2,9 +2,10 @@
 
 UUID_FIELD_NAME = 'crowdanki_uuid'
 DECK_FILE_NAME = "deck"
-SETTINGS_FIELD_NAME = '_settings'
 DECK_FILE_EXTENSION = ".json"
 MEDIA_SUBDIRECTORY_NAME = "media"
+IMPORT_CONFIG_NAME = "import_config"
+CONFIG_EXTENSION = ".yaml"
```
Stvad, line 8: not sure if splitting these adds much value tbh

ohare93: :shrug: don't much mind

ohare93:
@Stvad any idea when you'll have time to give me some feedback in the realms of "remove/add X and Y and I will gladly pull this" 😉

Stvad:
@ohare93 apologies for the delay :( I commit to do so by the end of the week (please poke me if I don't do so by Sat evening)

ohare93:
Sounds good mate, thanks :+1:

crowd_anki/importer/anki_importer.py
```diff
@@ -62,21 +54,59 @@ def load_from_directory(self, directory_path, import_media=True):
...
+        # TODO: strip settings from deck_json
+
+        return deck_json, import_dialog.final_import_config
+
     @staticmethod
-    def import_deck_from_path(collection, directory_path, import_media=True):
+    def read_deck(file_path):
+        """
+        Load deck from json file
+        :type file_path: Path
```
Stvad, line 83: just use the proper python type annotations instead

crowd_anki/importer/anki_importer.py
```diff
@@ -62,21 +54,59 @@ def load_from_directory(self, directory_path, import_media=True):
...
+    def import_deck_from_path(collection, directory_path):
         importer = AnkiJsonImporter(collection)
         try:
-            importer.load_from_directory(directory_path, import_media)
-            aqt.utils.showInfo("Import of {} deck was successful".format(directory_path.name))
+            deck_json, import_config = importer.load_deck_with_settings(directory_path)
+
+            if deck_json is not None:
+                importer.load_deck(deck_json, directory_path, import_config=import_config)
+                aqt.utils.showInfo("Import of {} deck was successful".format(directory_path.name))
```
Stvad, line 109: the way this is structured right now does not make sense to me. I think it should be something like:  
* have a  top level function like `load_deck` that gets the directory path
* internally it loads config (function) and gets update from the user (function)
* then it loads json and saves it to the collection (function)
* then it loads media if required (another function)

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,142 @@
+from collections import namedtuple
+from dataclasses import dataclass
+from enum import Enum
+
+from aqt.qt import *
+from typing import List, Tuple
+
+from .import_ui import Ui_Dialog as ConfigUI
+# from .config_settings import ConfigSettings
```
Stvad, line 9: 👀

crowd_anki/representation/json_serializable.py
```diff
@@ -6,9 +6,12 @@
...
 class JsonSerializable:
     readable_names = {}
-    export_filter_set = {"mod",  # Modification time
-                         "usn",  # Todo clarify
-                         "id"}
+    export_filter_set = {
+        "mod",  # Modification time
+        "usn",  # Todo clarify
+        "id",
+        "import_config"
```
Stvad, line 13: why do we need this?

crowd_anki/importer/anki_importer.py
ohare93: I swear I copied this from somewhere else in your code :thinking: but I can no longer find it. Anyways, changed :sweat_smile:

crowd_anki/importer/anki_importer.py
ohare93: It was structured this was as the import_config was inside the json file, so I had to read that first. I believe I can restructure it as stated :+1:

crowd_anki/importer/import_dialog.py
ohare93: Tried some crazy shenanigans :grin: removed

crowd_anki/representation/json_serializable.py
ohare93: Because the import config (the final one that is created at the end of the user import selection window) is added to the deck, and then passed down to the notes and whatnot. This is so that each can check it for their own desired import option, such as notes which look for the personal_field settings.

crowd_anki/importer/anki_importer.py
ohare93: Oh no, nevermind. I need to read in the deck in order to populate the note models as Personal Field options.

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class ImportConfig:
+    personal_fields: List[Tuple[str, str]] = []
+    add_tag_to_cards: List[str] = []
+
+    use_header: bool = True
+    use_note_models: bool = True
+    use_notes: bool = True
+    use_media: bool = True
```
Stvad, line 43: without the dataclass annotation these would be the class variables not the object ones, which is probably what you intended

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+
+from .import_ui import Ui_Dialog as ConfigUI
+from ..config.config_settings import ConfigSettings
+
+from ..utils.utils import string_cs_to_list
+
+ConfigEntry = namedtuple("ConfigEntry", ["config_name", "default_value"])
+
+
+class ImportJsonSetting:
```
Stvad, line 16: this name confuses me

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def __init__(self, settings_dict: dict):
+        for prop in self.Properties:
+            setattr(self, prop.value.config_name, settings_dict.get(prop.value.config_name, prop.value.default_value))
+
+        models = settings_dict.get("note_models", [])
+        for model_name in models:
+            personal_fields = models[model_name].get("personal_fields", [])
+            for pf in personal_fields:
+                self.personal_fields.append((model_name, pf))
```
Stvad, line 33: separate function

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+            self.form.cb_ignore_move_cards.setCheckState(Qt.Checked)
+
+        self.form.cb_headers.setCheckState(Qt.Checked)
+        self.form.cb_note_models.setCheckState(Qt.Checked)
+        self.form.cb_notes.setCheckState(Qt.Checked)
+        self.form.cb_media.setCheckState(Qt.Checked)
+
+        # TODO: Deck Parts to Use
+
+    def get_import_config(self):
```
Stvad, line 123: the name is misleading as we don't return anything. maybe `read_import_config` or something

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+        config = ImportConfig()
+
+        config.personal_fields = []
+        current_note_model = ""
+        fields: List[QListWidgetItem] = [self.form.list_personal_fields.item(i) for i in
+                                         range(self.form.list_personal_fields.count())]
```
Stvad, line 129: separate function

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+        for field in fields:
```
Stvad, line 130: maybe split things into different sub-panels instead of having a continuous list of heterogeneous  widgets?

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+        self.form.cb_note_models.setCheckState(Qt.Checked)
+        self.form.cb_notes.setCheckState(Qt.Checked)
+        self.form.cb_media.setCheckState(Qt.Checked)
+
+        # TODO: Deck Parts to Use
+
+    def get_import_config(self):
+        config = ImportConfig()
+
+        config.personal_fields = []
```
Stvad, line 126: extract parsing of the persona fields into a separate function

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self.form.cb_headers.setCheckState(Qt.Checked)
+        self.form.cb_note_models.setCheckState(Qt.Checked)
+        self.form.cb_notes.setCheckState(Qt.Checked)
+        self.form.cb_media.setCheckState(Qt.Checked)
+
+        # TODO: Deck Parts to Use
+
+    def get_import_config(self):
+        config = ImportConfig()
```
Stvad, line 124: consider using constructor

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+        self.ui_initial_setup()
+
+        self.final_import_config: ImportConfig = None
+
+    def accept(self):
+        self.get_import_config()
+        super().accept()
+
+    def reject(self):
+        self.get_import_config()
```
Stvad, line 71: do we need to do this on `reject`?

crowd_anki/representation/note.py
```diff
@@ -116,12 +118,28 @@ def save_to_collection(self, collection, deck, model_map_cache):
...
         self.anki_object.mod = anki.utils.intTime()
         self.anki_object.flush()
 
         if new_note:
             collection.addNote(self.anki_object)
-        elif not self.config.import_notes_ignore_deck_movement:
+        elif not self.import_config.ignore_deck_movement:
             self.move_cards_to_deck(deck.anki_dict["id"])
+
+    def handle_dictionary_update(self, note_model):
```
Stvad, line 131: `update_note_dictionary`?

crowd_anki/representation/note.py
```diff
@@ -116,12 +118,28 @@ def save_to_collection(self, collection, deck, model_map_cache):
...
+
+        # Personal Fields Resolution
+        fields = self.anki_object_dict["fields"]
```
Stvad, line 134: `imported_field_values`?

crowd_anki/representation/note.py
```diff
@@ -116,12 +118,28 @@ def save_to_collection(self, collection, deck, model_map_cache):
...
 
         if new_note:
             collection.addNote(self.anki_object)
-        elif not self.config.import_notes_ignore_deck_movement:
+        elif not self.import_config.ignore_deck_movement:
             self.move_cards_to_deck(deck.anki_dict["id"])
+
+    def handle_dictionary_update(self, note_model):
+
+        # Personal Fields Resolution
```
Stvad, line 133: extract into a separate function

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+
+        models = settings_dict.get("note_models", [])
+        for model_name in models:
+            personal_fields = models[model_name].get("personal_fields", [])
+            for pf in personal_fields:
+                self.personal_fields.append((model_name, pf))
+
+
+class ImportConfig:
+    personal_fields: List[Tuple[str, str]] = []
```
Stvad, line 37: why not make it a dict?

crowd_anki/representation/json_serializable.py
ohare93: Ahh, yes you are correct! :grin: I am only using the import_config values inside the individual save_to_collection functions anyways, so this is a much better solution. Changed :+1:

crowd_anki/representation/note.py
ohare93: Made a `handle_import_config_changes` function for both Personal Fields and Tags changes, and got rid of the useless variables. Have a look, I think it's more clear and streamlined :+1:

crowd_anki/representation/note.py
ohare93: See above.

crowd_anki/representation/note.py
ohare93: Changed

crowd_anki/utils/constants.py
ohare93: Removed.

ohare93:
Alright, just went ahead and implemented the Toggles for `Use Notes` and `Use Note Models`, as they were rather easy to do. Note Models I had to pass the bool into the `save_to_collection` function so that I could do part of it, to update the anki_dict, as the id is used in the notes importer. But notes was a simple if statement :smile: 

I gave implementing `Use Headers` (You call this the Deck and Metadata) but since it is split up into two segments, which are quite interlinked to all the others it I gave up for now. Also if the user has unticked `Use Headers` then perhaps they should also be forced to choose an existing deck for the new cards to go in :thinking: or just put them all in `Default`? What do you think? :thinking: 

Anyways, that means all the Deck Part Toggles apart from `Use Headers` are now functional too :+1: :grin:

ohare93:
> Speaking of future work, I am considering adding a counter for the amount of Notes, Note Models, and Media files being imported into this window. Was running tests using a deck with 57,000 notes, and realised I had no idea it has that many even as the person importing it scream Thoughts? thinking Would be trivial to implement.

Alright, too much time on my hands :laughing: so I implemented the count feature too!

image

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+
+from aqt.qt import *
+from typing import List, Tuple
+
+from .import_ui import Ui_Dialog as ConfigUI
+from ..config.config_settings import ConfigSettings
+
+from ..utils.utils import string_cs_to_list
+
+ConfigEntry = namedtuple("ConfigEntry", ["config_name", "default_value"])
```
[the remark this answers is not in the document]
Stvad: better type information/auto-completion/etc

crowd_anki/representation/deck.py
```diff
@@ -137,7 +139,7 @@ def save_to_collection(self,
 
         if save_note_models:
             for note_model in self.metadata.models.values():
-                note_model.save_to_collection(collection)
+                note_model.save_to_collection(collection, import_config.use_note_models)
```
Stvad, line 142: um, aren't `save_note_model` and `import_config.use_note_models` identical in intent? if so shouldn't the latter replace the former?

crowd_anki/importer/import_dialog.py
```diff
@@ -143,10 +143,16 @@ def setup_misc(self):
...
+        def set_checked_and_text(checkbox, text, count=0, checked: bool = True):
+            checkbox.setCheckState(Qt.Checked if checked else Qt.Unchecked)
+            if count:
+                text = f"{text}: {'{:,}'.format(count)}"
+            checkbox.setText(text)
+
+        set_checked_and_text(self.form.cb_headers, "Headers")
+        set_checked_and_text(self.form.cb_note_models, "Note Models", len(self.deck_json['note_models']))
+        set_checked_and_text(self.form.cb_notes, f"Notes", len(self.deck_json['notes']))
+        set_checked_and_text(self.form.cb_media, f"Media Files", len(self.deck_json['media_files']))
```
Stvad: separate function for these? :)

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,146 @@
...
+    use_header: bool = True
+    use_note_models: bool = True
+    use_notes: bool = True
+    use_media: bool = True
+
+    ignore_deck_movement: bool = False
+
+    def __repr__(self):
+        return f"ImportConfig({self.personal_fields!r}, {self.add_tag_to_cards!r}, " \
+               f"{self.use_header!r}, {self.use_notes!r}, {self.use_note_models!r}, {self.use_media!r} " \
```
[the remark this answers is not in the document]
Stvad: here it's just primitives, so I don't think we need that, but thanks for the reference! 🙂

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+
+@dataclass
+class ConfigEntry:
+    config_name: str
+    default_value: any
+
+
+@dataclass
+class PersonalFieldsHolder:
+    personal_fields: Dict[str, List[str]] = field(init=False, default_factory=lambda: dict())
```
Stvad, line 23: nit can just be `default_factory=dict`

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def has_pf(self, field_name, *keys):
```
Stvad, line 25: don't do abbreviations. this should probably be `contains_field`

Stvad: also should `field_name`  be `model_name` actually?

Stvad: oh wait, this checks if any of the models given contains the specified personal field. and we need it because we support both uids and the names now?  
I think supporting just the names is fine they should be unique withing given export

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+                    return True
+            return False
+        return any(_check_pfs(key) for key in keys)
+
+    def add_field(self, model_name, model_id, field_name):
+        def _add(key):
+            if key not in self.personal_fields:
+                self.personal_fields.setdefault(key, [field_name])
+            else:
+                self.personal_fields[key].append(field_name)
```
Stvad, line 38: look into `defaultdict`

crowd_anki/representation/note.py
Stvad: I'd still extract them into separate functions - each time you have a block comment explaining what the given subsection of code does - you'd be be better of extracting it into a separate function with self-explanatory name.  
Really recommend https://simpleprogrammer.com/respecting-abstraction/ as a read on general principles

crowd_anki/importer/import_dialog.py
Stvad: on looking how it's used, I think the interface should be something like `is_personal_field(model, field_name`

crowd_anki/representation/deck_initializer.py
```diff
@@ -4,6 +4,7 @@
 from .note import Note
 from ..anki.adapters.anki_deck import AnkiDeck
 from ..anki.adapters.note_model_file_provider import NoteModelFileProvider
+from ..importer.import_dialog import ImportConfig
```
Stvad, line 7: 👀

crowd_anki/representation/note_model.py
```diff
@@ -33,7 +33,7 @@ def check_semantically_identical(first_model, second_model):
 
         return True
 
-    def save_to_collection(self, collection: Collection):
+    def save_to_collection(self, collection: Collection, should_save: bool):
```
Stvad, line 36: https://github.com/Stvad/CrowdAnki/pull/72/files#r446706208

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+        ui_fields: List[QListWidgetItem] = self.get_ui_pf_items()
+        for ui_field in ui_fields:
+            if not ui_field.flags() & Qt.ItemIsUserCheckable:  # Note Model Header
+                current_note_model = ui_field.text()
+                current_uuid = next(model[UUID_FIELD_NAME]
+                                    for model in self.deck_json["note_models"] if model["name"] == current_note_model)
+            elif ui_field.checkState() == Qt.Checked:  # Field
+                config.add_field(current_note_model, current_uuid, ui_field.text())
+
+    def get_ui_pf_items(self):
```
Stvad, line 188: as elsewhere use full `personal_field`

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+        models = settings_dict.get("note_models", [])
+        for model_name_or_id in models:
+            self.personal_fields.setdefault(model_name_or_id, models[model_name_or_id].get("personal_fields", []))
+
+
+@dataclass
+class ImportConfig(PersonalFieldsHolder):
+    add_tag_to_cards: List[str]
+
+    use_header: bool
```
Stvad, line 70: btw what is header?

Stvad: >I gave implementing Use Headers (You call this the Deck and Metadata) a go, but since it is split up into two segments, which are quite interlinked to all the others, I gave up for now. Also if the user has unticked Use Headers then perhaps they should also be forced to choose an existing deck for the new cards to go in 🤔 or just put them all in Default? What do you think? 🤔

ah, you mention that it's not implemented for now. I believe it should be removed until it is. I'm still confused on what do you mean by header though?

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+            ignore_deck_movement=self.form.cb_ignore_move_cards.isChecked()
+        )
+
+        self.read_personal_fields(config)
+
+        self.final_import_config = config
+
+    def read_personal_fields(self, config):
+        current_note_model = ""
+        current_uuid = ""
```
Stvad, line 178: as mentioned elsewhere it's fined to just use name assuming it's unique

crowd_anki/importer/anki_importer.py
```diff
@@ -6,77 +6,100 @@
...
+
+        if not file_path.exists():
+            import_dict = {}
+        else:
+            with file_path.open(encoding='utf8') as meta_file:
+                import_dict = yaml.full_load(meta_file)
+
+        import_dialog = ImportDialog(deck_json, import_dict)
+        if import_dialog.exec_() == QDialog.Rejected:
+            return None, False
```
Stvad, line 93: probably ok to just return `None` and check for that

crowd_anki/importer/import_dialog.py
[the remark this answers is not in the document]
agentydragon: is the custom __repr__ needed in the first place? looks to me like it's generated for dataclasses by default.

https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+
+        def set_checked_and_text(checkbox, text, count, checked: bool = True):
+            checkbox.setCheckState(Qt.Checked if checked else Qt.Unchecked)
+            if count is not None:
+                text = f"{text}: {'{:,}'.format(count)}"
+            checkbox.setText(text)
+
+        set_checked_and_text(self.form.cb_headers, "Headers", None)
+        set_checked_and_text(self.form.cb_note_models, "Note Models", len(self.deck_json['note_models']))
+        set_checked_and_text(self.form.cb_notes, f"Notes", len(self.deck_json['notes']))
```
agentydragon, line 154: nit: f-string not needed

crowd_anki/representation/note.py
```diff
@@ -116,12 +117,23 @@ def save_to_collection(self, collection, deck, model_map_cache):
...
 
         if new_note:
             collection.addNote(self.anki_object)
-        elif not self.config.import_notes_ignore_deck_movement:
+        elif not import_config.ignore_deck_movement:
             self.move_cards_to_deck(deck.anki_dict["id"])
+
+    def handle_import_config_changes(self, import_config, note_model):
+        # Personal Fields
+        for num, field in enumerate(self.anki_object_dict["fields"]):
```
agentydragon, line 134: nit: the field var is never used, rename to _ (which is usually used for "throwaway vars")

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+        self.ui_initial_setup()
+
+        self.final_import_config: ImportConfig = None
+
+    def accept(self):
+        self.read_import_config()
+        super().accept()
+
+    def reject(self):
+        super().reject()
```
agentydragon, line 101: I think you can remove the reject method.

crowd_anki/importer/import_dialog.py
```diff
@@ -0,0 +1,189 @@
...
+        self.read_personal_fields(config)
+
+        self.final_import_config = config
+
+    def read_personal_fields(self, config):
+        current_note_model = ""
+        current_uuid = ""
+        ui_fields: List[QListWidgetItem] = self.get_ui_pf_items()
+        for ui_field in ui_fields:
+            if not ui_field.flags() & Qt.ItemIsUserCheckable:  # Note Model Header
```
agentydragon, line 181: could you save those 2 fields into some variable instead of cycling over all fields like this to find them?

crowd_anki/importer/import_dialog.py
agentydragon: Deck Owner?

crowd_anki/representation/note_model.py
ohare93: See my other comment on `save_note_model` not being replaced. Also this example further shows the need to pass it into here, as I need to call *part* of this function in order to keep everything else happy and working. I did elaborate in one of my top level comments in the PR:

> Note Models I had to pass the bool into the save_to_collection function so that I could do part of it, to update the anki_dict, as the id is used in the notes importer.

crowd_anki/importer/import_dialog.py
Stvad: eh, I still like the "Curator" best, owner has different connotations for me :p. how about "Maintainer" ?

ohare93:
Maintainer! Now that's a nice catch all that doesn't necessarily require it to be the owner of the repo. I swing my vote behind maintainer 😁

On Sat, 4 Jul 2020, at 8:31 PM, Vladyslav Sitalo wrote:
> 

> *@Stvad* commented on this pull request.

> 
> In crowd_anki/importer/import_dialog.py <https://github.com/Stvad/CrowdAnki/pull/72#discussion_r449795300>:

> > +            heading_ui = QListWidgetItem(model["name"])
+            heading_ui.setFlags(Qt.ItemIsEnabled)
+            self.form.list_personal_fields.addItem(heading_ui)
+
+            for field in model["flds"]:
+                field_ui = QListWidgetItem(field["name"])
+                field_ui.setCheckState(Qt.Checked if is_field_personal(field) else Qt.Unchecked)
+                field_ui.setFlags(Qt.ItemIsEnabled | Qt.ItemIsUserCheckable)
+                self.form.list_personal_fields.addItem(field_ui)
+
+    def setup_misc(self):
+        self.form.import_message_textbox.setText(self.import_settings.import_message)
+
+        if self.import_settings.suggest_tag_imported_cards:
+            self.form.cb_tag_cards.setCheckState(Qt.Checked)
+            self.form.cb_tag_cards.setText("Tag Cards (Suggested by Deck Manager!)")

crowd_anki/importer/import_dialog.py
```diff
@@ -149,10 +136,12 @@ def set_checked_and_text(checkbox, text, count, checked: bool = True):
                 text = f"{text}: {'{:,}'.format(count)}"
             checkbox.setText(text)
 
-        set_checked_and_text(self.form.cb_headers, "Headers", None)
+        self.form.cb_headers.setVisible(False)  # TODO: implement header selection functionality. Disabled for now
+        # set_checked_and_text(self.form.cb_headers, "Headers", None)
```
Stvad, line 140: just remove it for now ;)

crowd_anki/importer/import_dialog.py
Stvad: I'm not sure if it's useful to combine all those things together. I find it confusing, I think I won't be alone.   
Deck options can be separated out and it makes sense to do so #31   
But CrowdAnki is build on the notion of the Deck, I'd say it wont be straightforward to diverse it from the notion of a deck and I'm not sure if it makes sense to do so either

crowd_anki/representation/deck.py
Stvad: the metadata is shared between whole deck hierarchy. I'm still confused why these are different, let's chat when you're online

crowd_anki/importer/import_dialog.py
ohare93: :sob: my babies :sob: okay done :+1: :grin:

crowd_anki/importer/import_dialog.py
ohare93: See reply here https://github.com/Stvad/CrowdAnki/issues/31#issuecomment-653851949 :wink:

crowd_anki/representation/deck.py
ohare93: So you suggested the following in our WhatsApp chat:

> I think it'd make sense to read the config when you call the function from outside the deck and pass in save_to_collection as read in by import_config.

Tbh I have no idea what you mean, even looking at the code. Gonna need some clarification there :sweat_smile: 

*However!* In an attempt to tidy up this code in my own way I have extrapolated out the saving children and notes (the recursive part) into a separate function, and gotten rid of most of the predefined variables littering it up. Take a look at the new system and let me know how you like it. I have tested a few imports with child decks and it works fine :+1:

crowd_anki/importer/import_dialog.py
ohare93: :shrug: Done

crowd_anki/representation/deck.py
```diff
@@ -126,30 +126,27 @@ def _load_metadata_from_json(self, json_dict):
...
+        for note_model in self.metadata.models.values():
+            note_model.save_to_collection(collection)
+
+        self.save_children_and_notes(collection=collection,
+                                     import_config=import_config,
+                                     parent_name="",
+                                     model_map_cache=defaultdict(dict))
+
+    def save_children_and_notes(self, collection, import_config: ImportConfig, parent_name, model_map_cache):
         name = self._save_deck(collection, parent_name)
```
Stvad, line 142: this should still be part of main function though

crowd_anki/representation/deck.py
Stvad: The new structure makes sense!

crowd_anki/representation/deck.py
Stvad: ah, it does not quite work in that case. the method name is misleading in this case though, as it saves the deck itself as-well

ohare93: Yes it does also save the deck, hmm. Alright, just changed the name to `save_deck_children_and_notes` to be more clear. If you have a better suggestion lemme know :+1:

ohare93:
Merged and release?! :exploding_head: :partying_face: Fantastic, thank you for your many many reviews @Stvad :clap: we got there in the end :grin: :tada:

crowd_anki/representation/deck.py
ohare93: Oh nevermind, you already fixed it :laughing:

Issue #46, opened 2019-03-29
Import deck from git repository link


evolverine:
Instead of just downloading the repository, collaboration on decks requires cloning repositories, in order to have the original remote configured, so that people can push their changes to github.

Stvad:
The original design was caused by the fact that there was no actual git integration present. Now with dulwich being one of the dependencies this should be fairly easy to implement.

**Some requirements:**

* Clone into the snapshot directory
* If repository already exists in snapshot directory
    * Do `git remote add` (if not present)
    * `git pull <new remote> master`

evolverine:
working on this now, should have a pull request ready in a couple of days.

Stvad:
nice! Thank you!

PR #78, opened 2020-03-25
#46 git import

evolverine:
A few notes:

- you can ignore the first four commits. I made them when I wasn't yet aware that Dulwich was already a dependency of this project, and I was including GitPython. This is reverted in d502c48
- commit e72ef09 introduces a change in the directory structure that the repository is created in: instead of "CrowdAnki/repo_name/git" it uses "CrowdAnki/user_name/repo_name/". This is reverted in the last commits, but I do want to ask you about it, because I think it's much better to use the github username in the path, to avoid naming conflicts. What do you think, will this also impact on other parts of the add-on?
- the feature we have now is that we clone a repository instead of downloading the zip (and pull if the repository exists). I have not yet implemented the remote addition, but I'm happy to do it next if the direction seems good.

Also, two questions:

1. Instead of the Anki profile directory, shouldn't we now use the directory in the plugin settings to clone the repository?
2. For everyone but the repository maintainers the initially-cloned repository will have to be converted into a fork if they want to contribute any changes. Maybe we can automate that within this add-on using [hub](https://hub.github.com/)?


crowd_anki/utils/uuid.py
```diff
@@ -0,0 +1,597 @@
+r"""UUID objects (universally unique identifiers) according to RFC 4122.
```
Stvad: err, what's this about/for?

crowd_anki/github/github_importer.py
```diff
@@ -1,18 +1,14 @@
...
 from ..importer.anki_importer import AnkiJsonImporter
-from ..utils import utils
 
 BRANCH_NAME = "master"
-GITHUB_LINK = "https://github.com/{}/archive/" + BRANCH_NAME + ".zip"
+GITHUB_ZIP_URL = "https://github.com/{}/archive/" + BRANCH_NAME + ".zip"
+GITHUB_REPO_URL = "https://github.com/{}.git"
 
 
 class GithubImporter(object):
```
Stvad: should we rename this to GitImporter and accept just the repo URL?

crowd_anki/github/github_importer.py
```diff
@@ -32,24 +28,15 @@ def import_from_github(self):
         repo, ok = QInputDialog.getText(None, 'Enter GitHub repository',
                                         'Path:', text='<name>/<repository>')
         if repo and ok:
-            self.download_and_import(repo)
+            self.download_and_import_git(repo)
```
Stvad: why rename?

crowd_anki/github/github_importer.py
```diff
@@ -32,24 +28,15 @@ def import_from_github(self):
...
 
-    def download_and_import(self, repo):
+    def download_and_import_git(self, github_repo):
+        repo_parts = github_repo.split("/")
+        repo_dir = Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", repo_parts[-1], "git")
```
Stvad: For the directory I think the ideal option would be to clone things into the place where snapshots go (it's available in config).  The format there is `snapshot_dir/<anki profile name>/<deck name>`

Deck name is a bit tricky - if you follow the convention of naming the repo with the CrowdAnki deck name like here https://github.com/Stvad/Software_Engineering__git then you can use that, but I suspect you don't (being not a fun of those names & all :p)

So another option can be is storing the snapshot destination on per deck basis. Something like:
* If you're importing deck from git repo -> set snapshot destination to the original import directory
* If it's not set -> set it to the default one 
* Use these names on snapshot

crowd_anki/github/github_importer.py
```diff
@@ -32,24 +28,15 @@ def import_from_github(self):
...
-            utils.fs_remove(deck_directory)
-            deck_directory_wb.rename(deck_directory)
-            # Todo progressbar on download
-
-            AnkiJsonImporter.import_deck_from_path(self.collection, deck_directory)
-
-        except (URLError, HTTPError, OSError) as error:
-            aqt.utils.showWarning("Error while trying to get deck from GitHub: {}".format(error))
-            raise
+            repo_dir.mkdir(parents=True, exist_ok=True)
```
Stvad: why do we do mkdir in the pull branch? presumably it should be in the clone branch?

crowd_anki/github/github_importer.py
```diff
@@ -32,24 +28,15 @@ def import_from_github(self):
...
+            porcelain.pull(porcelain.open_repo(str(repo_dir)), GITHUB_REPO_URL.format(github_repo))
+        except NotGitRepository:  # Clone repository
+            porcelain.clone(GITHUB_REPO_URL.format(github_repo), repo_dir, False, True, porcelain.NoneStream(), porcelain.NoneStream())
```
Stvad: us named parameters here, otherwise this looks confusing

crowd_anki/github/github_importer.py
```diff
@@ -1,18 +1,14 @@
...
 from aqt import QInputDialog
+
+from dulwich import porcelain
+from dulwich.errors import NotGitRepository
 from ..importer.anki_importer import AnkiJsonImporter
-from ..utils import utils
 
 BRANCH_NAME = "master"
-GITHUB_LINK = "https://github.com/{}/archive/" + BRANCH_NAME + ".zip"
+GITHUB_ZIP_URL = "https://github.com/{}/archive/" + BRANCH_NAME + ".zip"
```
Stvad: I imagine this is not needed anymore

Stvad:
Integrating with github to create forks/etc - In general seems like a good idea for simplifying the flow, we should discuss how interaction would look like before going there though. And should be a separate CR.

Also using the actual cli tool really makes life harder distribution-wise, I imagine there is python lib out there for talking to Github API, which would be a better choice

crowd_anki/github/github_importer.py
evolverine: I'd say sure.

crowd_anki/github/github_importer.py
evolverine: Good point. It was because "download and import" before used to mean "download zip and import". So now it could be renamed to "clone_and_import", perhaps?

crowd_anki/utils/uuid.py
evolverine: No idea! Seems to have been a copy of [this](https://docs.python.org/3/library/uuid.html), but I don't know how it got there! It's out now.

evolverine:
agreed, a python library would be great for that.

crowd_anki/github/github_importer.py
evolverine: I'll admit that I didn't understand very well what you meant by the three bullet points above. Please see the commits I just pushed, and how they relate to what you suggest, thank you.

crowd_anki/github/github_importer.py
evolverine: Done

crowd_anki/github/github_importer.py
evolverine: Have you had a chance to take a look? Thanks.

Stvad: Should be able to take a closer look today/tomorrow.   

What the comment above was getting at that if the name of repository does not match the name of the deck - then initial import would work but subsequent snapshots will go to a different directory which is problematic. I proposed a way to solve that by changing how we determine the snapshot directory for deck

crowd_anki/config/config_settings.py
```diff
@@ -49,17 +49,26 @@ class Properties(Enum):
         EXPORT_CREATE_DECK_SUBDIRECTORY = ConfigEntry("export_create_deck_subdirectory", True)
         IMPORT_NOTES_IGNORE_DECK_MOVEMENT = ConfigEntry("import_notes_ignore_deck_movement", False)
 
-    def __init__(self, addon_manager=None, init_values=None):
+    def __init__(self, addon_manager=None, init_values=None, window=None):
         self.addon_manager = addon_manager or mw.addonManager
         self._config = init_values or addon_manager.getConfig(__name__)
+        self._window = window
```
Stvad, line 55: this should be initialized similar to the `addon_manage` (i.e. be assigned `mw` by default) also probably should be a more specific dependency - does `pm` stands for something like `profile_manager`?

crowd_anki/config/config_settings.py
```diff
@@ -49,17 +49,26 @@ class Properties(Enum):
...
-            cls.__instance = ConfigSettings(addon_manager)
+            cls.__instance = ConfigSettings(addon_manager=addon_manager, window=window)
         return cls.__instance
 
+    @property
+    def profileName(self):
+        return self._window.pm.name if self._window.pm else ""
+
+    @property
+    def snapshot_complete_dir(self):
```
Stvad, line 69: how about `full_snapshot_path` + why return a string instead of a `Path`?

crowd_anki/github/github_importer.py
```diff
@@ -1,40 +1,42 @@
...
-            self.clone_and_import(repo)
+            self.clone_repository_and_import(repo)
 
-    def clone_and_import(self, github_repo):
-        repo_parts = github_repo.split("/")
-        repo_dir = Path(self.collection.media.dir()).joinpath("..", "CrowdAnkiGit", repo_parts[-1], "git")
+    def clone_repository_and_import(self, repo_url):
+        repo_name = repo_url.split("/")[-1].split(".")[0]
+        repo_path = Path(ConfigSettings.get_instance().snapshot_complete_dir).joinpath(repo_name)
+        repo_dir = str(repo_path)
```
Stvad, line 31: extract into a separate function

crowd_anki/github/github_importer.py
```diff
@@ -1,40 +1,42 @@
...
         try:
-            porcelain.pull(porcelain.open_repo(str(repo_dir)), GITHUB_REPO_URL.format(github_repo))
+            porcelain.pull(porcelain.open_repo(repo_dir), repo_url)
         except NotGitRepository:  # Clone repository
-            repo_dir.mkdir(parents=True, exist_ok=True)
-            porcelain.clone(GITHUB_REPO_URL.format(github_repo), target=repo_dir, bare=False, checkout=True, errstream=porcelain.NoneStream(), outstream=porcelain.NoneStream())
-
-        AnkiJsonImporter.import_deck_from_path(self.collection, repo_dir)
+            try:
```
Stvad, line 35: this should be a separate function

crowd_anki/github/github_importer.py
```diff
@@ -1,40 +1,42 @@
...
+                repo_path.mkdir(parents=True, exist_ok=True)
+                porcelain.clone(repo_url, target=repo_dir, bare=False, checkout=True, errstream=porcelain.NoneStream(),
+                    outstream=porcelain.NoneStream())
+            except GitProtocolError as error: # git repository not found at that URL; but not sure how to display a more user-friendly error message
```
Stvad, line 39: see `AnkiUiNotifier` class and it's usages

katrinleinweber:
> […] python lib […] for talking to Github API

> […] would be great for that.

Have you found [developer.github.com/v3/libraries](https://developer.github.com/v3/libraries/)? It has a `Python` section, of which [PyGithub](https://github.com/PyGithub/PyGithub#pygithub) seems to be quite API-complete :-)

crowd_anki/config/config_settings.py
evolverine: Do you mean like this? (see the latest commit)

crowd_anki/github/github_importer.py
evolverine: But the name of the repository does not determine the name of the deck: when importing from github, the deck name is taken from `deck.json` So even if I import [the "covid-19-anki-deck](https://github.com/evolverine/covid-19-anki-deck/), the deck I see in Anki is "COVID-19" (in fact, I see "COVID-19_2", "COVID-19_3", etc, but right now that's beyond the point). Indeed, though, the name of the imported repository directory will be `covid-19-anki-deck`, so I see your point.

Nevertheless, the user can rename the deck to something wildly different, and they can also move the cards in and out of it at will. Being driven by the deck name will lead to a multitude of issues.

The way it would make the most sense for me for all this to happen is if each deck had a property called `remote_url` or something like that, which pointed to (either the original, or to the forked) remote repository. Then we'd use this property to determine the directory names for the repositories.
I'm not familiar enough with the Anki codebase to know whether that's something that could be implemented?

evolverine:
Thanks @katrinleinweber for researching the libraries. Can't wait to have this implemented. For now, let's finalize this pull request.

evolverine:
@Stvad did you have a chance to take a look at the commits? Thanks.

Stvad:
@evolverine I apologize for the delay, I thought I did, but I missed the latest batch. I'll take a look in the next few days!

crowd_anki/config/config_settings.py
```diff
@@ -49,17 +49,27 @@ class Properties(Enum):
         EXPORT_CREATE_DECK_SUBDIRECTORY = ConfigEntry("export_create_deck_subdirectory", True)
         IMPORT_NOTES_IGNORE_DECK_MOVEMENT = ConfigEntry("import_notes_ignore_deck_movement", False)
 
-    def __init__(self, addon_manager=None, init_values=None):
+    def __init__(self, addon_manager=None, init_values=None, window=None):
+        window = window or mw
```
Stvad, line 53: some nits here:   
* pass in the `profile_manager` instead of window, as we only need it
* order `addon_manager` and `profile_manager` to be adjacent in parameter list (as they are conceptually similar vs `init_values`)

Stvad:
Looks good to me besides the small nit that I've noted!

evolverine:
great. And check out the last commit as well, where I made improvements in the way the repository URL is parsed. If that looks good, on my side it's all done (otherwise let me know if any changes are in order).

crowd_anki/github/github_importer.py
```diff
@@ -8,6 +8,16 @@
 
 BRANCH_NAME = "master"
 
+
+def get_repository_name(repository_url):
+    repo_url_path = urlparse(repository_url).path
```
Stvad, line 13: what does using urlparse give us?

crowd_anki/github/github_importer.py
```diff
@@ -8,6 +8,16 @@
...
+    repo_name = list(filter(None, repo_url_path.split("/")))[-1]  # using filter() in case it ends in "/"
+    repo_name = repo_name.split(".")[0]  # in case it ends in ".git"
+    if repo_name.find(" ") != -1:
+        raise ValueError("repository name in URL should not contain spaces")
```
Stvad, line 17: that's seems like an oddly specific thing to check for. should we just strip the string?

crowd_anki/github/github_importer.py
evolverine: true, not much. I thought it could help in cases when the URL contains query parameters, but nothing in github seems to.

crowd_anki/github/github_importer.py
evolverine: Took it out as well. I think the reason I was trying to validate the input so much was that Dulwich does create a folder and empty repository even when the URL is not correct, or does not contain a git repository. But we can handle that in other ways, perhaps.

evolverine:
Heya, did you have a chance to look at the latest changes? Thanks!

Stvad:
Hey, thank you for a reminder, been a bit swamped recently =\. Will take a look today

Stvad:
Looks good! Thank you for working on this! Sorry for a slow turn-around

evolverine:
Yaay! Calls for a celebration 😃️🎉️

Issue #75, opened 2020-01-23
Collaboration improvement requests by Lukas Mertens


Stvad:
Relevant: CrowdAnki diff-merge tool: https://gitlab.com/dodod/crowdanki-diff

- [ ] Enable GUID sorting by default to minimize merge conflits at the end of the notes array in JSON (related to #71)
- [ ] Automatically trigger `check media` functionality on import to re-generate LaTeX images (https://apps.ankiweb.net/docs/manual.html#latex-on-web/mobile)
- [ ] Potentially filter out `newlyAdded` field from notes (https://github.com/ankitects/anki/blob/master/pylib/anki/notes.py#L23)
- [ ]  Things related to partial export also discussed in #74 and #62
- [ ] Note model merge is being too trigger happy - fires on structurally identical models

ohare93:
I fully support all of this 😁 I may give this crowdanki diff tool a try soon.

lukas-mertens:
Hey, CrowdAnki-Diff developer here: That would be very nice! Please note the current _early alpha everything could explode and kill you_ state of the project though. We would be very happy about some help, because I am currently very busy and not able to work on CrowdAnki-Diff for probably the next month or so. If someone could make some pull request, that would be greatly appreciated! Bug reports are welcome as well, of course!

Another thing I thought about was forking [Git-Json-Merge](https://github.com/jonatanpedersen/git-json-merge) for crowdanki. Maybe that could be a nice solution as well, as CrowdAnki-Diff only shows the differences, but doesn't produce a merged result yet.

lukas-mertens:
BTW: If you want to try out CrowdAnki-Diff you can try it on https://gitlab.com/dodod/karteikarten btw. We are currently actively working with it in that repo and we are trying to establish a working collaboration-workflow. We are still far from perfect, but from time to time it actually works quite smooth. It's a german repository, but probably you can use google translate to understand what we are doing. Otherwise feel free to ask. The README documents a lot though. If you want to see CrowdAnki-Diff in action, you can just click on one of the links that our bot posts to all merge requests.

ukanuk:
>  Note model merge is being too trigger happy - fires on structurally identical models

I've noticed it triggers whenever CSS or card templates change, even if the fields have not changed.

PR #82, opened 2020-04-15
New Note Filter for "newlyAdded"

ohare93:
Was comparing some exports and saw there's a new field on Notes called "newlyAdded". Having a look at the Anki source it seems it's just used for notification purposes, just now at least. Don't think this is wanted, or needed, in a CrowdAnkiExport though.

Added it into the Note Filter. May be good to get some type of automated test to check this in the future :thinking:


Stvad:
yeah, mentioned it in #75

ohare93:
Ah yes, so you did :grin:

PR #85, opened 2020-05-24
Hotfix for Import on Latest Anki

### LINKED ISSUE #84: Compatibility issues with the new version of Anki
Hey @Stvad ! I love and appreciate the work you put into this addon. I am having an issue with importing decks from any source (disk or github). Please let me know what I can do to fix this. Here is the error:
image

Here are my addons:
image

ohare93:
Working Version: Fix UUID get, and remove "mod" and "usn" from export filter. Fixes #84 

Ran into this issue while working on #72, so I decided to take a crack at it. This little fix makes things work again, but I believe it is not a great solution :slightly_frowning_face: 

The 'DeckManager' object has no attribute 'dconf' issue seemed to be caused by some restructuring of Decks/Deck Managers in Anki, so I've got it to point to the right place ("decks" instead of "dconf"). After this was fixed another issue cropped up, complaining there were missing properties. It seems that the latest Anki requires "mod" and "usn" on import, for some reason :thinking: I removed them from the export_filter, re-exported my test deck, and imported it successfully :+1:

Though I do not like the idea of mod and usn being in the json export files, further cluttering things up. Would be better for CrowdAnki to insert the mod and usn values. According to [this documentation](https://github.com/ankidroid/Anki-Android/wiki/Database-Structure) the mod is just the modification time (doable) but the usn may be more tricky :thinking: "-1" is probably fine, but may cause some sync issues?

Anyways I am unsure what the perfect fix for this is, or whether it should be fixed on Anki's end, so here's this working PR for now. Pull it, reject it, brainstorms other ideas and I can give them a try if I have time :+1:

### DIFF
--- a/crowd_anki/anki/overrides/uuid_operations.py
+++ b/crowd_anki/anki/overrides/uuid_operations.py
@@ -34,7 +34,7 @@ def get_deck_by_uuid(self, uuid):
 
 # Deck configuration
 def get_deck_configuration_by_uuid(self, uuid):
-    return get_from_dict_by_uuid(self, "dconf", uuid)
+    return get_from_dict_by_uuid(self, "decks", uuid)
 
 
 # Note model
--- a/crowd_anki/representation/json_serializable.py
+++ b/crowd_anki/representation/json_serializable.py
@@ -6,8 +6,8 @@ from ..utils.constants import UUID_FIELD_NAME
 
 class JsonSerializable:
     readable_names = {}
-    export_filter_set = {"mod",  # Modification time
-                         "usn",  # Todo clarify
+    export_filter_set = {  # "mod",  # Modification time
+                           # "usn",  # Todo clarify
                          "id"}
     import_filter_set = {"__type__"}
 


Stvad:
Hey! @ohare93  really appreciate you looking into this one!

crowd_anki/anki/overrides/uuid_operations.py
```diff
@@ -34,7 +34,7 @@ def get_deck_by_uuid(self, uuid):
 
 # Deck configuration
 def get_deck_configuration_by_uuid(self, uuid):
-    return get_from_dict_by_uuid(self, "dconf", uuid)
+    return get_from_dict_by_uuid(self, "decks", uuid)
```
Stvad, line 37: I don't think it's the right one. if you see above - we already query this field for decks, here we need deck config

crowd_anki/representation/json_serializable.py
```diff
@@ -6,8 +6,8 @@
 
 class JsonSerializable:
     readable_names = {}
-    export_filter_set = {"mod",  # Modification time
-                         "usn",  # Todo clarify
+    export_filter_set = {  # "mod",  # Modification time
+                           # "usn",  # Todo clarify
```
Stvad, line 10: hrm, previously Anki would set those to defaults if they were missing. I supposed we can just steal code from previous version =\

Stvad: actually this didn't break for me for some reason?

Stvad:
Closing this in favor of #86 Would appreciate your help with testing that one @ohare93

crowd_anki/anki/overrides/uuid_operations.py
ohare93: Couldn't see what was supposed to be there before, just saw a CrowdAnki_UUID right there next to where it was trying to fetch when stopping with a breakpoint :man_shrugging: tried it out of desperation, which is why I labelled mine as a hotfix :sweat_smile:

crowd_anki/representation/json_serializable.py
ohare93: Must've been my fault then.

PR #91, opened 2020-07-06
Personal Fields Tidy Up

ohare93:
Fix Unit Test Imports and UI Layout/Wording


Stvad:
Can you post a screenshot? :) From WhatsApp, I remember it was somewhat misaligned still (if it's that version)

ohare93:
It is the same one you have seen, yes.

image

I think the old version looked strange since the top right hand column occupied the same space as the header of the top left column "Import Message". This one may have a smidge too much space from the top, but I'd say it's better than the current version :man_shrugging: 

Since you fixed all the other issues this PR is just about UI spacing :sweat_smile: so it's not so important

Stvad:
I agree that this is better but why not something like (with elements on the right pushed to the top vs bottom):  
86712311-2a086800-c01d-11ea-9324-ae8ae89034ef

Issue #98, opened 2020-09-27
Error when specifying Deck Sort Methods


xiaoxiae:
Hello,

I have a particular CrowdAnki setup, where a snapshot of the deck is automatically created on closing Anki and commited to Git. This has worked wonders for the past year, but when opening it today, I noticed that the order of the items in JSON keeps changing as the snapshots are created.

This is obviously a huge problem, so I tried to remedy this by using TODO in Tools > Add-ons > CrowdAnki > Deck Sort Methods. However, changing the text in any way results in the following error:

```
Caught exception:
Traceback (most recent call last):
  File "/home/xiaoxiae/Documents/Anki/addons21/1788670778/config/config_dialog.py", line 77, in changed_textedit_deck_sort_methods
    self.config.export_note_sort_methods = self.string_cs_to_list(
AttributeError: 'ConfigDialog' object has no attribute 'string_cs_to_list'
```

Am I doing something wrong?

xiaoxiae:
It seems that removing `self.` before `string_cs_to_list` works (since `string_cs_to_list` is a function imported from `..utils.utils`), but it seems that it has no effect and the attributes are still sorted quite randomly.

aplaice:
Thanks very much for diagnosing the miscalling of `string_cs_to_list` — it indeed fixes the crash!

However, with the fix, I can't reproduce the issue of the snapshots not being sorted.

Both `field1` and `guid` as the "deck sort method" seemed to work as expected. I tried it with Anki 2.1.26, 2.1.28 and 2.1.34.

Which sort method and which Anki version are you using?

(Thanks for reporting!)

xiaoxiae:
I tried using both to no avail, but maybe I just don't understand what it's supposed to do. I expect there to be no diff when I create two snapshots, one after another, without changing my Anki deck. Using `guid` and `field1` doesn't seem to do this.

aplaice:
Yes, there should be a diff, when changing the sorting method! (Except if your deck happens to be already sorted by both fields, but that's obviously unlikely enough to be effectively impossible.)

ohare93:
@xiaoxiae do you mind sharing what you are exporting? Just the json is fine. Then we can inform you whether the export happened successfully, and things were sorted by guid then field1 👍 A screenshot of your config settings window just before successfully saving would also go a long way. Thanks mate.

xiaoxiae:
I don't mind at all. Using Anki 2.1.26 and the following config:
```
[...]

PR #102, opened 2020-10-09
Fix crash when editing deck sort method

aplaice:
Credit goes to @xiaoxiae ([98#issuecomment-699616688](https://github.com/Stvad/CrowdAnki/issues/98#issuecomment-699616688))!


Stvad:
Thanks @aplaice, @xiaoxiae

ohare93:
Thanks @aplaice :clap: I entirely forgot about that part of the issue :scream: good thing it was an easy fix! :sweat_smile:

PR #114, opened 2020-12-18
Update exporter to work with Anki ≥ 2.1.36

### LINKED ISSUE #113: Export function not working.
When I try to export, I get the following error:

```Error 
If the issue only occurs when add-ons are enabled, please use the Tools>Add-ons menu item to disable some add-ons and restart Anki, repeat until you discover the add-on that is causing the problem. 
When you've discovered the add-on that is causing the problem, please report the issue on the add-ons section of our support site. 
Debug info:
Anki 2.1.37 (6d596c8f) Python 3.8.6 Qt 5.14.2 PyQt 5.14.2
Platform: Windows 10
Flags: frz=True ao=True sv=1
Add-ons, last update check: 2020-12-18 15:34:33

Caught exception:
  File "aqt\main.py", line 1114, in onExport
  File "aqt\exporting.py", line 33, in __init__
  File "aqt\exporting.py", line 48, in setup
  File "<decorator-gen-2>", line 2, in exporterChanged
  File "anki\hooks.py", line 89, in decorator_wrapper
  File "anki\hooks.py", line 81, in repl
  File "C:\Users\Username\AppData\Roaming\Anki2\addons21\1788670778\anki\overrides\exporting.py", line 17, in exporter_changed
    self.exporter = aqt.exporting.exporters()[exporter_id][1](self.col)
TypeError: exporters() missing 1 required positional argument: 'col'
```

The only addon I have installed is CrowdAnki.

aplaice:
`exporters` now takes an argument.

See [e3b4802f47395b9c1a75ff89505410e91f34477e](https://github.com/ankitects/anki/commit/e3b4802f47395b9c1a75ff89505410e91f34477e#diff-3ec9206f4c144d89f905bfafebf91de634efc7f344b611b4fcbe9ec880ad39c5R446).

The commit was after 2.1.35 but before 2.1.36.  (Testing explicitly shows that this change is needed for 2.1.36 and 2.1.37, but breaks 2.1.35.)

Fix #113.

### DIFF
--- a/crowd_anki/anki/overrides/exporting.py
+++ b/crowd_anki/anki/overrides/exporting.py
@@ -14,7 +14,7 @@ from ...utils import constants
 
 
 def exporter_changed(self, exporter_id):
-    self.exporter = aqt.exporting.exporters()[exporter_id][1](self.col)
+    self.exporter = aqt.exporting.exporters(self.col)[exporter_id][1](self.col)
     self.frm.includeMedia.setVisible(hasattr(self.exporter, "includeMedia"))
 
 


aplaice:
AFAICT this is the only breaking change since 2.1.35 — I very briefly tested a simple export, snapshot and import, and they all seemed to work.

Stvad:
thanks!
*sigh* let me try to upload this in a way that does not break things...

Stvad:
done

Issue #118, opened 2021-02-05
Options group created from another options group inherits its uuid (minor issue)


aplaice:
If you use Anki's interface for creating new options groups (or "deck configs" in CrowdAnki terms), by clicking on the cog next to a deck, selecting "Options" > "Options group" > "Manage" > "Add"\*, then the newly created options group will be a "clone" of the previously selected options group.

\* I think that this is the only way to create an additional options group, in Anki, without add-ons.

In particular, this means that the newly created "deck config" will inherit/share the `crowdanki_uuid` of its "parent" "deck config".

### To reproduce

1. Create and open a new Anki profile.
2. Create a snapshot. ("File" > "CrowdAnki: Snapshot"). (Alternatively, you can use a CrowdAnki export of the Default deck, but snapshots are less clicks :)). (If you wish, you can create some notes first, but it's not necessary.)
3. Inspect the `deck_config_uuid` (or alternatively the `crowdanki_uuid` in `deck_configurations`) of the snapshotted deck.
3. Create a new options group (via clicking on the cog next to the Default deck > "Options" > "Options group" > "Manage" > "Add") and set it as the options group for the Default deck.
4. Create a snapshot.
5. Inspect the `deck_config_uuid` of the snapshotted deck.

(Tested with Anki 2.1.39, on Linux.)

### Expected behaviour

The two `deck_config_uuid`s are different.

### Actual behaviour

The two `deck_config_uuid`s are the same.

#### Alternative order

If you perform the actions in a different order — i.e. create a second options group and switch back to the first options group before the first snapshot, and switch to the already created second options group after the first snapshot, then the `deck_config_uuid`s will be different.

(Full alternative order:

1. Create and open a new Anki profile.
2. Create a new options group (via clicking on the cog next to the Default deck > "Options" > "Options group" > "Manage" > "Add"), but switch back to the original ("Default") options group.
3. Create a snapshot. ("File" > "CrowdAnki: Snapshot"). (Alternatively, you can use a CrowdAnki export of the Default deck, but snapshots are less clicks :)). (If you wish, you can create some notes first, but it's not necessary.)
4. Inspect the `deck_config_uuid` (or alternatively the `crowdanki_uuid` in `deck_configurations`) of the snapshotted deck.
5. Switch to the options group created in step 2.
4. Create a snapshot.
5. Inspect the `deck_config_uuid` of the snapshotted deck.

)

[...]

PR #127, opened 2021-04-27
Avoid storing `deck_config_uuid` in Anki db and deal with fallout

### LINKED ISSUE #61: Deck won't round-trip.
Hi, I've got a deck that won't round-trip. It seems fine in Anki, but when I export it and then import it, it gives the error below: 

I'm happy to help you debug it! What do you need?

Error 
If the issue only occurs when add-ons are enabled, please use the Tools>Add-ons menu item to disable some add-ons and restart Anki, repeating until you discover the add-on that is causing the problem. 
When you've discovered the add-on that is causing the problem, please report the issue on the add-ons section of our support site. 
Debug info:
Anki 2.1.11 (3cf770c7) Python 3.6.7 Qt 5.12.1 PyQt 5.11.3
Platform: Linux
Flags: frz=True ao=True sv=1
Add-ons possibly involved: CrowdAnki JSON exportimport Edit history Collaborate on deck creation

Caught exception:
  File "/home/john/.local/share/Anki2/addons21/1788670778/anki/ui/action_vendor.py", line 29, in <lambda>
    lambda: AnkiJsonImporter.import_deck(self.window.col, self.directory_vendor))
  File "/home/john/.local/share/Anki2/addons21/1788670778/importer/anki_importer.py", line 80, in import_deck
    AnkiJsonImporter.import_deck_from_path(collection, Path(directory_path))
  File "/home/john/.local/share/Anki2/addons21/1788670778/importer/anki_importer.py", line 69, in import_deck_from_path
    importer.load_from_directory(directory_path, import_media)
  File "/home/john/.local/share/Anki2/addons21/1788670778/importer/anki_importer.py", line 45, in load_from_directory
    self.load_from_file(directory_path.joinpath(directory_path.name).with_suffix(DECK_FILE_EXTENSION))
  File "/home/john/.local/share/Anki2/addons21/1788670778/importer/anki_importer.py", line 30, in load_from_file
    deck.save_to_collection(self.collection)
  File "/home/john/.local/share/Anki2/addons21/1788670778/representation/deck.py", line 141, in save_to_collection
    name = self._save_deck(collection, parent_name)
  File "/home/john/.local/share/Anki2/addons21/1788670778/representation/deck.py", line 171, in _save_deck
    self.anki_dict["conf"] = self.metadata.deck_configs[self.deck_config_uuid].anki_dict["id"]
<class 'KeyError'>: 'c675e2b3-9a7b-11e9-9bae-342387e2ac49'

### LINKED ISSUE #106: Cant import from github
Hello!, I cant import the deck I last used 10 months ago (jan2020)

here is the repo:
https://github.com/joecabezas/k014_anki_deck_00_99_major_system_pegs

I just modified the repo so there is a single json with the same name of the repo (expecting this to fix my issue but not)

Anki version: 2.1.35 (latest)

repro steps:

* install Anki 2.1.35
* add addon: 1788670778
* restart Anki
* import from github repo: https://github.com/joecabezas/k014_anki_deck_00_99_major_system_pegs

it shows error:
```python
Error
If the issue only occurs when add-ons are enabled, please use the Tools > Add-ons menu item to disable some add-ons and restart Anki, repeating until you discover the add-on that is causing the problem.
When you've discovered the add-on that is causing the problem, please report the issue on the add-on support site.
Debug info:
Anki 2.1.35 (84dcaa86) Python 3.8.0 Qt 5.14.2 PyQt 5.14.2
Platform: Windows 10
Flags: frz=True ao=True sv=1
Add-ons, last update check: 2020-10-27 23:27:33

Caught exception:
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\anki\ui\action_vendor.py", line 35, in <lambda>
    lambda: GitImporter.on_git_import_action(self.window.col))
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\github\github_importer.py", line 28, in on_git_import_action
    GitImporter(collection).import_from_git()
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\github\github_importer.py", line 33, in import_from_git
    self.clone_repository_and_import(repo_url)
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\github\github_importer.py", line 48, in clone_repository_and_import
    AnkiJsonImporter.import_deck_from_path(self.collection, repo_local_path)
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\importer\anki_importer.py", line 101, in import_deck_from_path
    if importer.load_deck(directory_path):
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\importer\anki_importer.py", line 39, in load_deck
    deck.save_to_collection(self.collection, import_config=import_config)
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\representation\deck.py", line 131, in save_to_collection
    self.save_decks_and_notes(collection=collection,
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\representation\deck.py", line 144, in save_decks_and_notes
    full_name = self._save_deck(collection, parent_name)
  File "C:\Users\Joe\AppData\Roaming\Anki2\addons21\1788670778\representation\deck.py", line 173, in _save_deck
    self.anki_dict["conf"] = self.metadata.deck_configs[self.deck_config_uuid].anki_dict["id"]
KeyError: '8cc79fcf-35d9-11ea-9a85-305a3a05e8e9'
```

I tried downgrading to 2.1.26 as stated in https://github.com/Stvad/CrowdAnki/issues/96
Also tested with Anki 2.1.25, Anki 2.1.15 with no luck

I cant make it to work again, I want to at least recover the deck, which is a single one, what I am missing?

-Joe

### LINKED ISSUE #116: Exporter retains deck config uuid despite not including config profile in exported JSON
Issue: as title.

To reproduce:
1. Have another person import an Anki deck not having your config profiles and then export it back to you.
2. Perform some edits and have them reimport your export.
3. The importer will throw an error saying it detected a nonexistent "deck_config_uuid"

Platform: Anki 2.1.35, Windows

137213513_704182567133602_8824641045323812466_n

aplaice:
Fix #106, fix #116, fix #61.

Due to the fact that the `deck_config_uuid` is imported by `CrowdAnki` into the Anki database as a property of the deck (in the `decks` table, at least in later versions of Anki), but not updated when the options group (Deck config) is changed, and due to the order of keys in `crowd_anki/representation/json_serializable.py` (`deck.anki_dict["deck_config_uuid"]` (with the stale/hardcoded `deck_config_uuid`) overrides `deck.deck_config_uuid` (which is fetched on demand, using `Deck._load_deck_config`) during export), a `deck.json` file with a deck whose `deck_config_uuid` points towards a Deck config that's not actually present in `deck_configurations` can be created. This obviously results in an error when a user tries to import the deck.

<hr/>

The reproduction steps are:

1. Start with a fresh profile.
2. Create a deck (say `foo`) with a couple of notes.
3. Create a new deck config (options group), say `new_options`, via the standard: `Options` > `Options group` > `Manage...` > `Add` and switch back to `Default`.

    Note that this has to be done at this stage, rather than later, otherwise we're confounded by #118.
    
4. Export `foo` with CrowdAnki.
5. Re-import `foo` with CrowdAnki.
6. Change the options group of `foo` from `Default` to `new_options`.
7. Export `foo` with CrowdAnki.
8. Try re-importing `foo`.

Expected result: All the imports and exports work fine.

Actual result: The final import fails with a `KeyError` (due to the mismatching `deck_config_uuid`).

<hr/>

I discussed some possible solutions [here](https://github.com/Stvad/CrowdAnki/issues/106#issuecomment-723312413).

After some thought, I think that the `deck_config_uuid` should simply not be stored in the Anki db (i.e. **2** from the linked comment). It duplicates the information present in the `conf` key in the db. [We already (sensibly) strip](https://github.com/Stvad/CrowdAnki/blob/117a2946aeeed00abf2b90430b70073636a66d3f/crowd_anki/representation/deck.py#L35) `note_models` or `children` from the deck object before writing it into the Anki db and there's no reason for `deck_config_uuid` to be different. (That's the first commit.)

Unfortunately, this will only prevent the issue from coming up in the future, for people with fresh Anki profiles and creating new decks. We'd still be left with:

1. Many deck maintainers/contributors with the `deck_config_uuid` still stored in the Anki database, attached to the Anki deck object.
   This means that if they ever change the options group of a deck with such a stale `deck_config_uuid`, and export it, we'll again have a broken `deck.json`.
2. There are probably broken exported `deck.json`s out there, which will crash users' Ankis if imported.
   IMO this is far less pressing than 1., because it's easy to notice and diagnose.

The second commit deals with 1. — it (permanently) removes the `deck_config_uuid` from the given deck's `anki_dict` when exporting that deck. (I thought about removing `deck_config_uuid`s from all decks, on `CrowdAnki` startup, but I think that that's too risky and invasive.) It can hopefully be removed in a year and so, once we're sure that we've cleared out all the `deck_config_uuid`s from (most) users' databases.

The third commit tries to deal with 2., by providing a modal popup with some info. I'm not sure if it's helpful or necessary, though.

<hr/>
    
    
We've been pretty lucky that we also have issue #118, as it's considerably mitigated #106/#116. #118 causes different deck configs to have the same `crowdanki_uuid`, so even though the `deck_config_uuid` is not updated when the deck config is changed, it doesn't matter, because both the "right" and the "wrong" `deck_config_uuid`s are the same. (In that it matters to the extent that the wrong deck config is probably exported, but at least import doesn't cause a crash.)

<hr/>

### Summary

Stale `deck_config_uuid` can cause broken `deck.json`s. (The `deck_config_uuid` is not among the `crowdanki_uuid`s of the deck configs in `deck_configurations`.)

1. `4ec62298` prevents the `deck_config_uuid` from being stored in the Anki database during import. (The `deck_config_uuid` duplicates the deck's `conf` and the deck config's `crowdanki_uuid`, so it's not really useful, but as noted above, can be harmful.)
2. `e8fa915f` removes the `deck_config_uuid` where it's already present in the database, during export.
3. `e8dca2f7` provides the end-user with some info if they do encounter a broken `deck.json` during import. I'm not sure if it actually adds any value.

**1** resolves the initial bug, **2** and **3** deal with the "fallout".

### DIFF
--- a/crowd_anki/representation/deck.py
+++ b/crowd_anki/representation/deck.py
@@ -83,6 +83,12 @@ class Deck(JsonSerializableAnkiDict):
         new_config = DeckConfig.from_collection(self.collection, self.anki_dict["conf"])
         self.deck_config_uuid = new_config.get_uuid()
 
+        # TODO Remove this once enough time has passed that #106/#116
+        # is no longer an issue — i.e. when there are likely no more
+        # Anki dbs with a `deck_config_uuid` attached to the deck.
+        if "deck_config_uuid" in self.anki_dict:
+            del self.anki_dict["deck_config_uuid"]
+
         self.metadata.deck_configs.setdefault(new_config.get_uuid(), new_config)
 
     def serialization_dict(self):
--- a/crowd_anki/representation/deck.py
+++ b/crowd_anki/representation/deck.py
@@ -87,6 +87,7 @@ class Deck(JsonSerializableAnkiDict):
         # TODO Remove this once enough time has passed that #106/#116
         # is no longer an issue — i.e. when there are likely no more
         # Anki dbs with a `deck_config_uuid` attached to the deck.
+        # See #133.
         if "deck_config_uuid" in self.anki_dict:
             del self.anki_dict["deck_config_uuid"]
 
@@ -184,6 +185,7 @@ class Deck(JsonSerializableAnkiDict):
             # TODO Remove the exception-handling once we're confident
             # that there are no more buggy decks, with mismatching
             # `deck_config_uuid`s.
+            # See #133.
         except KeyError as error:
             AnkiModalNotifier().error("Incorrect deck config",
                                       "The deck config uuid {} is not present in the deck. "
--- a/crowd_anki/representation/deck.py
+++ b/crowd_anki/representation/deck.py
@@ -36,7 +36,8 @@ class Deck(JsonSerializableAnkiDict):
                          "deck_configurations",
                          "children",
                          "media_files",
-                         "notes"}
+                         "notes",
+                         "deck_config_uuid"}
 
     def __init__(self,
                  file_provider_supplier: Callable[[Any, Iterable[int]], FileProvider],
--- a/crowd_anki/representation/deck.py
+++ b/crowd_anki/representation/deck.py
@@ -9,6 +9,7 @@ from ..importer.import_dialog import ImportConfig
 from ..utils import utils
 from ..utils.constants import UUID_FIELD_NAME
 from ..utils.uuid import UuidFetcher
+from ..utils.notifier import AnkiModalNotifier
 
 DeckMetadata = namedtuple("DeckMetadata", ["deck_configs", "models"])
 
@@ -177,7 +178,21 @@ class Deck(JsonSerializableAnkiDict):
 
         self.anki_dict = deck_dict
         self.anki_dict["name"] = full_name
-        self.anki_dict["conf"] = self.metadata.deck_configs[self.deck_config_uuid].anki_dict["id"]
+
+        try:
+            self.anki_dict["conf"] = self.metadata.deck_configs[self.deck_config_uuid].anki_dict["id"]
+            # TODO Remove the exception-handling once we're confident
+            # that there are no more buggy decks, with mismatching
+            # `deck_config_uuid`s.
+        except KeyError as error:
+            AnkiModalNotifier().error("Incorrect deck config",
+                                      "The deck config uuid {} is not present in the deck. "
+                                      "This is likely due to a now-fixed bug in CrowdAnki. "
+                                      "Please ask the maintainer of the deck to re-export it, if possible. "
+                                      "See here: https://github.com/Stvad/CrowdAnki/issues/106 "
+                                      "for details and alternative solutions.".format(error))
+            raise
+
         collection.decks.save(deck_dict)
 
         return full_name
--- a/crowd_anki/representation/deck_initializer.py
+++ b/crowd_anki/representation/deck_initializer.py
@@ -44,7 +44,6 @@ def from_json(json_dict, deck_metadata=None) -> Deck:
     deck.notes = [Note.from_json(json_note) for json_note in json_dict["notes"]]
     deck.children = [from_json(child, deck_metadata=deck.metadata) for child in json_dict["children"]]
 
-    # Todo should I call this here?
     deck.post_import_filter()
 
     return deck


crowd_anki/representation/deck.py
```diff
@@ -82,6 +84,12 @@ def _load_deck_config(self):
         new_config = DeckConfig.from_collection(self.collection, self.anki_dict["conf"])
         self.deck_config_uuid = new_config.get_uuid()
 
+        # TODO Remove this once enough time has passed that #106/#116
```
Stvad, line 87: create a separate issue to track this?

Stvad: also - does this actually help? would we write an updated version of the dict to db?

aplaice: > create a separate issue to track this and mention it in the comment?

Makes sense! Will do!

> also - does this actually help?

Yeah, it helps!

1. It ensures that the correct `deck_config_uuid` is set during that given export.

> would we write an updated version of the dict to db?

2. Yes, we do write an updated version of the deck dict to the db ([here](https://github.com/Stvad/CrowdAnki/blob/master/crowd_anki/export/anki_exporter.py#L71)), so we're safe going forward (even if we remove this code).

To double check *2*: 

(steps 1-4 are from adapted from [above](https://github.com/Stvad/CrowdAnki/pull/127#issue-624377195), but without the changing of the deck config, since we're only concerned with whether `deck_config_uuid` is present at all)

0. Use the current, published version of the add-on (or `master` from `git`).
1. Start with a fresh profile.
2. Create a deck (say `foo`) with a couple of notes.
3. Export `foo` with CrowdAnki.
4. Re-import `foo` with CrowdAnki.
5. Using the Anki console (accessed with <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>;</kbd>) check that `deck_config_uuid` is (undesirably) present in the Anki dict:

```python3
print('deck_config_uuid' in mw.col.decks.byName('foo'))
```
> `True`

6. Quit Anki, switch to the version of the add-on from this branch/PR and start Anki again.
7. (Optionally, do 5 again, getting `True`, again, to make sure that it's step 8 that removes `deck_config_uuid`, not restarting Anki etc.)
8. Export `foo` with CrowdAnki, again.
9. Like in 5 (and 7), check whether `deck_config_uuid` is present in the Anki dict for `foo`:
```python3
print('deck_config_uuid' in mw.col.decks.byName('foo'))
```
> False

PR #129, opened 2021-04-27
Prevent crash when trying to export all decks

### LINKED ISSUE #128: Trying to export all decks crashes Anki
### To reproduce:

1. Try to export all decks (`File` > `Export` > `CrowdAnki JSON representation` > `Export...`).

### Expected behaviour:

Export fails with an error message.

### Actual behaviour:

Anki crashes.

aplaice:
Fix #128.

The crash was caused by the warning dialog, itself!  The issue was that export is now run in a background thread, but GUI operations from a background thread cause a crash.

AnkiModalNotifier is used in `anki_exporter_wrapper.py` and `github_importer.py`.  The former is in a background thread, the latter in the main thread.  In both cases, error messages are now correctly displayed, without crashing.

Tested on Anki 2.1.40 and 2.1.43.

For comparison / a similar issue [see](https://anki.tenderapp.com/discussions/ankidesktop/42070-anki-closes-without-warning-when-importing-conflicting-shared-deck).

I'm not very happy about the solution — particularly the timer — but without the timer `run_on_main` results in some sort of weird deadlock and I can't think of a better method.

### DIFF
--- a/crowd_anki/utils/notifier.py
+++ b/crowd_anki/utils/notifier.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 import aqt.utils
+import aqt
 
 
 class Notifier(ABC):
@@ -16,16 +17,27 @@ class Notifier(ABC):
     def error(self, title: str, message: str):
         pass
 
+def run_closure_in_main(closure):
+    """Run the closure in the main thread.
+
+This is necessary, because GUI operations in a background thread cause
+a crash.  For instance, export is now run in a background thread.
+
+    """
+    if aqt.mw.inMainThread():
+        closure()
+    else:
+        aqt.mw.taskman.run_on_main(lambda: aqt.mw.progress.timer(50, closure, False))
 
 class AnkiModalNotifier(Notifier):
     def info(self, title: str, message: str):
-        aqt.utils.showInfo(message, title=title)
+        run_closure_in_main(lambda: aqt.utils.showInfo(message, title=title))
 
     def warning(self, title: str, message: str):
-        aqt.utils.showWarning(message, title=title)
+        run_closure_in_main(lambda: aqt.utils.showWarning(message, title=title))
 
     def error(self, title: str, message: str):
-        aqt.utils.showCritical(message, title=title)
+        run_closure_in_main(lambda: aqt.utils.showCritical(message, title=title))
 
 
 class AnkiTooltipNotifier(Notifier):
--- a/crowd_anki/utils/notifier.py
+++ b/crowd_anki/utils/notifier.py
@@ -27,7 +27,7 @@ a crash.  For instance, export is now run in a background thread.
     if aqt.mw.inMainThread():
         closure()
     else:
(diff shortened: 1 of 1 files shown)


crowd_anki/utils/notifier.py
```diff
@@ -16,16 +17,27 @@ def warning(self, title: str, message: str):
...
+    """Run the closure in the main thread.
+
+This is necessary, because GUI operations in a background thread cause
+a crash.  For instance, export is now run in a background thread.
+
+    """
+    if aqt.mw.inMainThread():
+        closure()
+    else:
+        aqt.mw.taskman.run_on_main(lambda: aqt.mw.progress.timer(50, closure, False))
```
Stvad, line 30: any reason not to have 0 there?

Stvad:
hey, apologies for the delay getting back to this. have some big changes in life that I'm focusing on. should be able to review both your PR's now though!

crowd_anki/utils/notifier.py
Stvad: or, actually what does that number signify? (I initially assumed - delay)

aplaice: It's the [initial delay time](https://github.com/ankitects/anki/blob/8555a977cbf4b3633eaeed328619c796aaba12b0/qt/aqt/progress.py#L32). If I understand correctly, Anki will try to run the closure after the initial delay (in ms) and then try repeatedly in 100 ms intervals, until it's possible to do so.

I picked 50 ms as small enough not to result in any (significant) visible latency, but large enough that 

I didn't use 0, since:

1. Anki itself doesn't seem to use 0 anywhere where it uses `mw.progress.timer`.
2. I thought (without testing the timing explicitly) that since Anki can't run the closure immediately (since without the timer Anki hangs), it's better to wait for 50 ms before trying for the first time, as then we might have a chance of succeeding. (With `0`, we try to run the closure at times (0, 100, 200, 300...), but presumably the first try will fail, so we'll probably end up running at 100 ms. With `50`, we try at (50, 150, 250...), but there's a chance the first time will actually succeed, so we might end up running at 50 ms, earlier than with `0`.)

However, explicitly testing the timing (with the following ugly make-shift contraption:

<details>

```python3
import time
# ...
    def warning(self, title: str, message: str):
        print(time.time())
        tttt = time.time()
        def foo():
            print(time.time())
            print(time.time() - tttt)
            aqt.utils.showWarning(message, title=title)
        run_closure_in_main(foo)
```

</details>

) suggests that using `0` is still faster. (For `50`, the times are roughly in 70-90 ms range, while for `0` they're in the 10-50 ms range. For `10`, they're in the 30-70 ms range.)

A timeout of 0 is explicitly considered in the [official Qt docs](https://doc.qt.io/qt-5/qtimer.html#details) valid, so it's not a weird edge-case that just accidentally works, so I'll switch to it!

aplaice:
No problem! Good luck with the life changes! Thanks for your review here!

crowd_anki/utils/notifier.py
Stvad: Thanks for diving deeper on this!

PR #135, opened 2021-07-03
Avoid duplicating crowdanki_uuid when cloning deck config

### LINKED ISSUE #118: Options group created from another options group inherits its uuid (minor issue)

aplaice:
Please do not merge this before at least the first commit of #127 (this bug (the #118 bug) mitigates the more serious bug that #127 solves)!

Partially deal with #118.

Testing, it seems that I somehow haven't broken the tests despite fiddling with anki/hook_vendor without adding to the overrides.

Note that `deck_conf_did_add_config` is not available as a "legacy" hook, so the new hook mechanism has to be used.  It might (or might not) be worth converting the existing hooks to the new mechanism.

I've chosen to simply delete the `crowdanki_uuid` property from the cloned deck config, rather than immediately regenerate a new uuid, to avoid code duplication — when the new deck config needs a uuid, one will be generated for it, by the standard mechanism.

### Testing

This PR can be tested to show that it works correctly by following the reproduction sequence in #118 or by inspecting the `crowdanki_uuid`s of the deck configs with the debug console:

1. Open the console with <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>;</kbd>.
2. Run the following and check that none of the UUIDs are the same (and are present/absent when expected):

```python3
for c in self.col.decks.all_config():
    if "crowdanki_uuid" in c:
        print(f"{c['crowdanki_uuid']} for {c['name']}")
    else:
        print(f"no uuid for {c['name']}")
```

### Duplicate deck config `crowdanki_uuid`s "in the wild"

There will almost certainly be many cases where users have deck configs with duplicate `crowdanki_uuid` and this PR obviously won't fix that — it'll just prevent more such cases occurring.  Since this is a minor bug, I don't think that it's worth complicating the CrowdAnki codebase by including an automated fixer.

Instead, I'll write a debug console script for diagnosing and fixing this and the other UUID issues.

### DIFF
--- a/crowd_anki/anki/hook_vendor.py
+++ b/crowd_anki/anki/hook_vendor.py
@@ -1,3 +1,5 @@
+from aqt import gui_hooks
+
 from dataclasses import dataclass
 from typing import Any
 
@@ -5,6 +7,7 @@ from ..config.config_settings import ConfigSettings
 from ..anki.adapters.hook_manager import AnkiHookManager
 from ..export.anki_exporter_wrapper import exporters_hook
 from ..history.archiver_vendor import ArchiverVendor
+from ..utils.deckconf import disambiguate_crowdanki_uuid
 
 
 @dataclass
@@ -16,6 +19,7 @@ class HookVendor:
     def setup_hooks(self):
         self.setup_exporter_hook()
         self.setup_snapshot_hooks()
+        self.setup_add_config_hook()
 
     def setup_exporter_hook(self):
         self.hook_manager.hook("exportersList", exporters_hook)
@@ -24,3 +28,6 @@ class HookVendor:
         snapshot_handler = ArchiverVendor(self.window, self.config).snapshot_on_sync
         self.hook_manager.hook('profileLoaded', snapshot_handler)
         self.hook_manager.hook('unloadProfile', snapshot_handler)
+
+    def setup_add_config_hook(self):
+        gui_hooks.deck_conf_did_add_config.append(disambiguate_crowdanki_uuid)
--- a/crowd_anki/utils/deckconf.py
+++ b/crowd_anki/utils/deckconf.py
@@ -1,0 +1,10 @@
+from .constants import UUID_FIELD_NAME
+
+def disambiguate_crowdanki_uuid(deck_conf, deck,
+                                config, new_name,
+                                new_conf_id):
+    new_deck_conf = deck_conf.mw.col.decks.get_config(new_conf_id)
+    if (new_deck_conf and (UUID_FIELD_NAME in new_deck_conf)):
+        # Delete rather than generating anew, (with uuid1()) to avoid code duplication.
+        del new_deck_conf[UUID_FIELD_NAME]
+        deck_conf.mw.col.decks.update_config(new_deck_conf)
--- a/crowd_anki/utils/deckconf.py
+++ b/crowd_anki/utils/deckconf.py
@@ -1,3 +1,7 @@
+# Note that this only works for Anki ≤ 2.1.45, or when using the old
+# deck options popup with Anki 2.1.46+.  Anki 2.1.46+ has a new
+# JavaScript-based options popup (with no hooks), by default.
+
 from .constants import UUID_FIELD_NAME
 
 def disambiguate_crowdanki_uuid(deck_conf, deck,


aplaice:
In Anki 2.1.46, the fix no longer works with the new deck options change pop-up (if one uses "Clone preset", the deck config's UUID is sadly also cloned). (The fix still works if one uses the old deck options Qt dialog, accessible with Shift-click, but that's little consolation, since it's sadly going away, in the long run.)

ohare93:
Just had this issue affect me, yet again :sob: I am on Anki 2.1.44 and cloned via the Clone Preset button, as described above.

Where are the CrowdAnkiUUIDs themselves stored? Could they be periodically (on start up or on command) all checked and duplicates changed?

What is the easy way to make a fix when in this broken state?

ohare93:
Fixed it a terrible manual way. Generated a new guid (replaced that in the CrowdAnki file), reimported back into Anki, then converted all the original notes to the new type and deleted the original note type. Really shitty that this is still happening though :disappointed: a regular user would have no chance to fix this, or even know what is going on

aplaice:
> Note type

Are you using CrowdAnki from `master` or from AnkiWeb? (The issue has been (mostly, subject to the provisos listed in #136\*) fixed in `master`, but not released yet, as Stvad is needed to upload to AnkiWeb and I don't want to bother him too frequently, so I wanted to get all the essential fixes into `master` before releasing.) (If you're on `master` and still being hit by the bug, then it's more problematic.)

\* The main proviso is that if you clone note types in an Anki/AnkiMobile/AnkiDroid instance other than the one where CrowdAnki is installed (and obviously you can't have CrowdAnki on mobile) you'll still be hit by the issue, but hopefully that should be rare. (Also, existing collections aren't cleaned up.)

> Where are the CrowdAnkiUUIDs themselves stored? Could they be periodically (on start up or on command) all checked and duplicates changed?

Yes, they could be periodically checked (to work around the proviso and to fix existing Anki collections), but it's hard to determine which note model id is the original and which the new one (in principle, `id`s should be monotonically increasing, but I think there are subtle edge cases when timestamps happen to clash, and I don't want to change the UUID for the old model).). (Alternatively, we could have a second (?) storage place for note_model_id to crowdanki_uuid lookups, which wouldn't be affected by Anki's model cloning.)

<hr/>

> Really shitty that this is still happening though

Yes, sorry!

<hr/>

(BTW this PR is about deck configs which almost nobody cares about, not note models.)

aplaice:
> What is the easy way to make a fix when in this broken state? [note: for note models not deck configs]

<details><summary>Debug console:</summary>

```python3
from uuid import uuid1

uuids = []

for model in filter(lambda model: 'crowdanki_uuid' in model,
                    sorted(self.col.models.all(), key=lambda m: m["id"])):
    # we're sorting in the hope that note models with higher ids
    # (which mostly corresponds to creation date) were created later,
    # so we change the uuid of the copies, not the original
    cu = model["crowdanki_uuid"]
    if cu in uuids:
        print("Replacing UUID for note model " + model["name"] + "!")
        model["crowdanki_uuid"] = str(uuid1())
        self.col.models.save(model)
    else:
        uuids.append(cu)

```

</details>

ohare93:
> > Really shitty that this is still happening though
> 
> Yes, sorry!
> 
> (BTW this PR is about deck configs which almost nobody cares about, not note models.)

No need to apologise :wink: I was complaining in general.


> > Note type
> 
> Are you using CrowdAnki from `master` or from AnkiWeb? (The issue has been (mostly, subject to the provisos listed in #136*) fixed in `master`, but not released yet, as Stvad is needed to upload to AnkiWeb and I don't want to bother him too frequently, so I wanted to get all the essential fixes into `master` before releasing.) (If you're on `master` and still being hit by the bug, then it's more problematic.)

I am on Master, but it seems I am out of date :sweat_smile: only 26 commits though :laughing: my own dumb fault, then


> > Where are the CrowdAnkiUUIDs themselves stored? Could they be periodically (on start up or on command) all checked and duplicates changed?
> 
> Yes, they could be periodically checked (to work around the proviso and to fix existing Anki collections), but it's hard to determine which note model id is the original and which the new one (in principle, `id`s should be monotonically increasing, but I think there are subtle edge cases when timestamps happen to clash, and I don't want to change the UUID for the old model).). (Alternatively, we could have a second (?) storage place for note_model_id to crowdanki_uuid lookups, which wouldn't be affected by Anki's model cloning.)
> 

Sounds like a reasonable fix :clap: I hope to have some time off soon, where I would be happy to help.

aplaice:
As noted, this fix doesn't help in Anki 2.1.46, with the new deck options change pop-up.  Unfortunately, the approach from this PR can't be adjusted for the new pop-up (due to its architecture — JS frontend communicating with the Rust backend via a _very_ thin Python wrapper).

Something like #155 would work, but it'd have to be the only solution (not just as a safety net as in #155), which doesn't feel very elegant.

However, the old deck options pop-up still appears when you use the old (v1) scheduler or shift-click on the "options" button (and obviously is fully functional for Anki 2.1.45 and lower), so I _think_ it's still worth including.

PR #137, opened 2021-07-12
Implement part of Github direct upload method

shaunakg:
This adds some methods into the exporter class and export wrapper class to facilitate the direct export of the JSON representation into Github. It also adds some config entry fields where the user can put their Github [personal access token](https://git.io/token), username and repository name.

This is not ready for merging yet because it does not actually show up in the UI as an export option. I am working on doing this but it will likely require the rewriting of part of the export dialog.

It includes the dependency `PyGithub` because it is much easier to use the API of Github over natively implementing Git through preexisting dependency`dulwich` and pushing the files. This might be able to be done but would be much more complicated (in my opinion).

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,6 +19,7 @@ dataclasses = "*"
 pygtrie = "*"
 cached-property = "*"
 PyYAML = "*"
+pygithub = "*"
 
 [requires]
 python_version = "3.7"
--- a/crowd_anki/config/config_dialog.py
+++ b/crowd_anki/config/config_dialog.py
@@ -35,11 +35,16 @@ class ConfigDialog(QDialog):
     def ui_initial_setup(self):
         self.setup_snapshot_options()
         self.setup_export_options()
+        self.setup_github_options()
         self.setup_import_options()
 
     def setup_snapshot_options(self):
         self.form.textedit_snapshot_path.setText(self.config.snapshot_path)
         self.form.textedit_snapshot_path.textChanged.connect(self.changed_textedit_snapshot_path)
+
+        self.form.textedit_gh_username.textChanged.connect(self.changed_gh_username)
+        self.form.textedit_gh_password.textChanged.connect(self.changed_gh_password)
+        self.form.textedit_gh_repo.textChanged.connect(self.changed_gh_repo)
 
         self.form.cb_automated_snapshot.setChecked(self.config.automated_snapshot)
         self.form.cb_automated_snapshot.stateChanged.connect(self.toggle_automated_snapshot)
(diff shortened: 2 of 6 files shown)


Stvad:
Hi @shaunakg,

thanks for contributing this!
I'm currently travelling and have reduced availability, but I should be ablet to look at this in ~1.5 weeks.  

re PyGithub - it's ok, as long as it does not transitively pull in dependency on native things (e.g. native git binary?)
---
@aplaice I'd appreciate you taking a look as-well if you have a chance!

aplaice:
This would be really convenient to have, so thanks very much for writing it!

Unfortunately, I haven't had a time to look at or test it.

> re PyGithub - it's ok, as long as it does not transitively pull in dependency on native things (e.g. native git binary?)

On a very cursory glance, it seems that `PyGithub` depends on `PyNaCl` which is a wrapper around `libsodium` — according to the [relevant documentation](https://pypi.org/project/PyNaCl/) "PyNaCl ships as a binary wheel on macOS, Windows and Linux `manylinux1`".  However, I haven't dug too deeply, so I'm hopefully wrong!

Stvad:
If it is bringing in a binary dependency - I don't know of a way to properly package it into the anki plugin such that it'd work properly on all platforms 🙁

ahmedyarub:
> If it is bringing in a binary dependency - I don't know of a way to properly package it into the anki plugin such that it'd work properly on all platforms 🙁

By "all platforms" do you mean mobile platforms also? In that case you can enable the feature only in Desktop platforms and disable it for the mobile ones.

aplaice:
> By "all platforms" do you mean mobile platforms also? In that case you can enable the feature only in Desktop platforms and disable it for the mobile ones.

No. The issue is that it seems that PyNaCl requires a compiled C component. 

The maintainers of PyNaCl have uploaded binary wheels to pypi, so that when installing with pip (or pipenv etc.), pip simply downloads the wheels from pypi, appropriate for the given platform (Windows, Linux, MacOS, etc.) and the given processor type (x86, arm(32|64) etc.), provided these wheels are available, so there's little issue.

However, an Anki add-on has to be self-contained, so for PyNaCl to function we'd have to include all the many different binary wheels and use the correct ones at run-time. The problems are:

1. I'm not sure if this (choosing among different packaged wheels at run-time) is even possible.
2. If it is, it's tricky to get right, and would increase complexity.
3. It'd likely considerably increase add-on size.
4. I'm not sure if uploading compiled code to AnkiWeb is even allowed.
5. We'd have to be very careful to include the binary wheels for all possible desktop platforms.

(Anki add-ons aren't available on mobile (iOS or Android) at all, so mobile Anki isn't the issue.)

ahmedyarub:
Thank you for the detailed reply. I'll try testing this on my machine but I remember a similar case of XGBoost where the binaries for all target platforms were included in the final package. I'll test and come back to you later.

aplaice:
An alternative might be to use a different python GitHub API interface — e.g. [ghapi](https://github.com/fastai/ghapi) — instead of [PyGitHub](https://github.com/PyGithub/PyGithub).

PR #160, opened 2021-12-17
Created a menu action to run create a snapshot and exit anki

evandrocoan:
It is taking some time to create my snapshot and it is not always desirable to create it when exiting/opening Anki.

### DIFF
--- a/crowd_anki/anki/ui/action_vendor.py
+++ b/crowd_anki/anki/ui/action_vendor.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass, field
 from typing import Callable, Any, Optional
+from aqt import mw
 
 from ...github.github_importer import GitImporter
 from ...history.archiver_vendor import ArchiverVendor
@@ -24,7 +25,7 @@ class ActionVendor:
         return action
 
     def actions(self):
-        return [self.import_action(), self.github_import(), self.snapshot()]
+        return [self.import_action(), self.github_import(), self.snapshot(), self.snapshot_and_exit()]
 
     def import_action(self):
         return self.action('CrowdAnki: Import from disk',
@@ -36,3 +37,10 @@ class ActionVendor:
 
     def snapshot(self):
         return self.action('CrowdAnki: Snapshot', self.archiver_vendor.do_manual_snapshot)
+
+    def _snapshot_and_exit(self):
+        self.archiver_vendor.do_manual_snapshot()
+        mw.close()
+
+    def snapshot_and_exit(self):
+        return self.action('CrowdAnki: Snapshot and Exit', self._snapshot_and_exit)


aplaice:
Thanks for this! It's a brilliant solution (or at least workaround) to the issue of snapshots taking a long time.

However, I'd prefer to be careful when adding extra menu items, since:

1. They "clutter up" the menu bar for all CrowdAnki users.

    CrowdAnki already takes up half of the space in the `File` menu.

2. They provide a suggestion of permanence to the given option. (Users might get used to having the menu action and even once we solve the issue of excruciatingly slow snapshots, they might be unhappy about it being removed.)

3. They might confuse new users. ("Import git repo", "Import from disk" and "Snapshot" are all pretty self-explanatory, but they might wonder why "Snapshot and exit" is needed.)

OTOH for users with a large collection, who snapshot, this new menu item might be a _huge_ "quality of life" improvement.

<hr/>

From what I understand, you run the add-on from your own private fork, anyway, so not merging this PR immediately won't significantly inconvenience you.(?)

Hence, I'll wait for (hopefully) some feedback from more people and think about how best to get the upsides of this PR without the (slight) downsides, for a while, before merging.

evandrocoan:
> but they might wonder why "Snapshot and exit" is needed

I could change it to something like `Snapshot and exit Anki (useful for big collections)`.

> From what I understand, you run the add-on from your own private fork

Yes. I in my fork, I added all files to the top-level (https://github.com/evandroforks/CrowdAnki/commit/e7e80c9de02b1ece0cfc58101884b2e09f985ffa) and directly added the PiPY dependencies into the fork code. This way, I can just clone the repository inside Anki addons directory, and the dependency will load directly from source-code.

I wonder how is the standard development workflow. Do you always build an Anki package and install it after changing the code? i.e., would the workflow be:
1. `bash package_plugin.sh`
2. `unzip packagefile.zip -d ~/Anki/addons21/`

If this would be the workflow, maybe add it README.md or some development file, as I usually run my Anki addons directly from source-code/git repository cloned directly to my addons directory.

aplaice:
> I could change it to something like Snapshot and exit Anki (useful for big collections).

That helps with **3** (explanation), but makes **1** worse, so I think the original wording is better.

<hr/>

> I wonder how is the standard development workflow. Do you always build an Anki package and install it after changing the code? i.e., would the workflow be:

The standard development workflow is [described in `CONTRIBUTING.md`](https://github.com/Stvad/CrowdAnki/blob/master/CONTRIBUTING.md#testing-your-changes-in-anki):

> 1. If you have the production version of CrowdAnki installed - remove it from Anki. 
> 1. Run `fetch_dependencies.sh` - this will download the dependencies required for CrowdAnki to 
operate and put them into the `crowd_anki/dist` directory.
> 1. Add a symlink to the Anki plugins directory (you can find it via `Tools>Add-ons>View Files`)
pointing to `crowd_anki` directory (**you should not** name the symlink `crowd_anki` though, see 
[#62](https://github.com/Stvad/CrowdAnki/issues/62)).

i.e. very close to your current workflow. (The difference is having a symlink vs. having the code directly in the relevant directory, but it's a cosmetic one.)

aplaice:
In the interest of "perfect" not getting in the way of "better", I'm merging this now.  Thanks again for contributing it and not just keeping it in your private fork!

<hr/>

I might eventually move this behind an option (i.e. displayed only if a given option is turned on) or into a submenu (e.g. `File > CrowdAnki (Advanced) > ...`). I'll let you know before I do so.

PR #167, opened 2022-05-15
Adapt to 2.1.50+ backup mechanism

### LINKED ISSUE #166: Anki 2.1.50 throws exception on crowdanki imports on macOS Monterey 12.3.1 (AttributeError: 'AnkiQt' object has no attribute 'backup')
Importing a crowdanki document on Anki 2.1.50 (File > Crowd Anki Import from Disk) causes an exception to be thrown the import to fail. Failure occurs with my existing Anki Profile or with a new testing Anki Profile. It fails with a variety of crowdanki documents.

Upgrading the latest crowdanki via the Anki menu,

Downgrading to Anki 2.1.49 works around the error.

Here's the copied error:

```
Error
If the issue only occurs when add-ons are enabled, please use the Tools > Add-ons menu item to disable some add-ons and restart Anki, repeating until you discover the add-on that is causing the problem.
When you've discovered the add-on that is causing the problem, please report the issue to the add-on author.
Debug info:
Anki 2.1.50 (26d40c3a) Python 3.9.7 Qt 6.2.3 PyQt 6.2.3
Platform: Mac 12.3.1
Flags: frz=True ao=True sv=?
Add-ons, last update check: 2022-04-23 16:03:32
Add-ons possibly involved: ⁨CrowdAnki JSON exportimport Edit history Collaborate on deck creation⁩

Caught exception:
  File "/Users/schwa/Library/Application Support/Anki2/addons21/1788670778/anki/ui/action_vendor.py", line 31, in <lambda>
    lambda: AnkiJsonImporter.import_deck(self.window.col, self.directory_vendor))
  File "/Users/schwa/Library/Application Support/Anki2/addons21/1788670778/importer/anki_importer.py", line 114, in import_deck
    AnkiJsonImporter.import_deck_from_path(collection, Path(directory_path))
  File "/Users/schwa/Library/Application Support/Anki2/addons21/1788670778/importer/anki_importer.py", line 103, in import_deck_from_path
    if importer.load_deck(directory_path):
  File "/Users/schwa/Library/Application Support/Anki2/addons21/1788670778/importer/anki_importer.py", line 37, in load_deck
    aqt.mw.backup()
AttributeError: 'AnkiQt' object has no attribute 'backup'
```

aplaice:
Fix #166.

It's no longer necessary for the collection to be closed before a backup, and AFAICT a closed collection can't even be backed up.

See PR 1728 in Anki itself.  (Not linking directly to avoid spam for DAE etc.)

This breaks compatibility with 2.1.49-.

Also: silence most verbose deprecation warnings (in 2.1.50+)

I've left the Qt warnings alone for now, since the behaviour varies between Qt5 and Qt6.

<hr/>

AFAICT importing, exporting and snapshotting all work in 2.1.50 and 2.1.52 both with Qt5 and Qt6 (using the default Qt5 compatibility shims).

### DIFF
--- a/crowd_anki/anki/overrides/exporting.py
+++ b/crowd_anki/anki/overrides/exporting.py
@@ -23,7 +23,7 @@ def get_save_file(parent, title, dir_description, key, ext, fname=None):
         directory = str(QFileDialog.getExistingDirectory(caption="Select Export Directory",
                                                          directory=fname))
         if directory:
-            return os.path.join(directory, str(anki.utils.intTime()))
+            return os.path.join(directory, str(anki.utils.int_time()))
         return None
 
     return aqt.utils.getSaveFile_old(parent, title, dir_description, key, ext, fname)
--- a/crowd_anki/importer/anki_importer.py
+++ b/crowd_anki/importer/anki_importer.py
@@ -33,9 +33,7 @@ class AnkiJsonImporter:
             return False
 
         if aqt.mw:
-            aqt.mw.col.close(downgrade=False)
-            aqt.mw.backup()
-            aqt.mw.col.reopen(after_full_sync=False)
+            aqt.mw.create_backup_now()
         try:
             deck = deck_initializer.from_json(deck_json)
             deck.save_to_collection(self.collection, import_config=import_config)
--- a/crowd_anki/representation/deck.py
+++ b/crowd_anki/representation/deck.py
@@ -107,7 +107,7 @@ class Deck(JsonSerializableAnkiDict):
             # TODO Remove compatibility shims for Anki 2.1.46 and
             # lower.
             join_fields = anki_object.joined_fields if hasattr(anki_object, 'joined_fields') else anki_object.joinedFields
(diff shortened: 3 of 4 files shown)


boydkelly:
> Fix #166.
> 
> It's no longer necessary for the collection to be closed before a backup, and AFAICT a closed collection can't even be backed up.
> 
> See PR 1728 in Anki itself. (Not linking directly to avoid spam for DAE etc.)
> 
> This breaks compatibility with 2.1.49-.
> 
> Also: silence most verbose deprecation warnings (in 2.1.50+)
> 
> I've left the Qt warnings alone for now, since the behaviour varies between Qt5 and Qt6.
> 
> AFAICT importing, exporting and snapshotting all work in 2.1.50 and 2.1.52 both with Qt5 and Qt6 (using the default Qt5 compatibility shims)al.

Tested also in 2.1.51 qt5 and 2.1.53 qt6   both imports successful.

Thank you very much!

aplaice:
Thanks very much for testing (especially 2.1.53)! (I'll try to check if there are any other regressions there, soon-ish.)

PR #175, opened 2022-11-06
Avoid mocking Anki modules globally

### LINKED ISSUE #145: Stop mocking Anki modules globally
(This is mostly to remind myself for the future.)

Currently, the presence of `mock_anki_modules()` in `test/config/config_settings_spec.py` causes the anki imports to be mocked, for all tests (including those that don't call `mock_anki_modules()`, when the tests are run together.

# To reproduce

## Steps

1. Install all dependencies (including dev):

```bash
pipenv install --dev
```

2. Run all mamba tests together:

```bash
pipenv run mamba ./
```

3. Run the tests in `test/export/note_sorter_spec.py` by themselves:

```bash
pipenv run mamba ./test/export/note_sorter_spec.py
```

## Expected result

Both for **2** and **3** all tests pass.

## Actual result

For **2** all tests pass:

```
31 examples ran (3 pending) in 1.2636 seconds
```

For **3** the tests crash:

<details>

```
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/bin/mamba", line 8, in <module>
    sys.exit(main())
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/mamba/cli.py", line 18, in main
    runner.run()
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/mamba/runners.py", line 29, in run
    modules = self.example_collector.modules()
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/mamba/example_collector.py", line 20, in modules
    with self._load_module_from(path) as module:
  File "/usr/local/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/mamba/example_collector.py", line 52, in _load_module_from
    yield self._module_from_ast(name, path)
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/mamba/example_collector.py", line 70, in _module_from_ast
    exec(code, module.__dict__)
  File "./test/export/note_sorter_spec.py", line 5, in <module>
    from aqt import mw
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/aqt/__init__.py", line 22, in <module>
    from aqt.qt import *
  File "/home/x/.local/share/virtualenvs/CrowdAnki-4rfLFX4y/lib/python3.7/site-packages/aqt/qt.py", line 16, in <module>
    from PyQt5.QtWebEngineWidgets import *  # type: ignore
ModuleNotFoundError: No module named 'PyQt5.QtWebEngineWidgets'
```

</details>

(The fact that the tests crash is another issue, but they should either always crash or always succeed, so the fact that their completion depends on the way they're called is problematic.)

# Alternative approach (to make sure)

## To reproduce

1. In `test/config/config_settings_spec.py` (the file containing the `mock_anki_modules()` call), comment out everything after `mock_anki_modules()`.  (We're not interested in the behaviour of `config_settings_spec.py` in the absence of module mocking.)

2. Run all mamba tests:

```bash
pipenv run mamba ./
```

3. Comment out `mock_anki_modules()` in `test/config/config_settings_spec.py`.

4. Run all mamba tests again.

## Expected result

The same result is returned in **2** and **4**.

## Actual result

For **2** the tests all succeed:

```
25 examples ran (3 pending) in 1.2638 seconds
```

For **4** the tests crashed again.

<hr/>

Presumably, given that `config` comes before `export` alphabetically, the patching of `sys.module` persists throughout the testing.

(I think that I haven't missed anything obvious, for why the behaviour is different.)

The main short-term conclusion (mainly for me) is that the behaviour of the tests is different, depending on whether they're called individually, or all together.

aplaice:
Ideally, we'd stop mocking them _at all_ but this is a good first step, as it's annoying and surprising (with the current/previous _global_ mocking, the test is/was run differently depending on whether it was run individually or as part of a batch).

Fix #145.

(The solution could probably be neater (e.g. we could use a context manager), but I don't think it's worth it.)

(The reproduction in #145 no longer works, as-is, as the tests don't crash even without mocking. (Edit: famous last words :D (though AFAICT the issue is PyQt5.QtWebEngineWidgets not wanting to be imported on Windows rather than a bug in our code).)

<hr/>

The issue is now `pywin32` which _might_ be tricky getting into the pipenv (even if we wanted to...). (Mocking it might be a simpler approach. :/)

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -304,6 +304,14 @@
             "markers": "sys_platform != 'darwin' and sys_platform != 'win32'",
             "version": "==1.5.0"
         },
+        "exceptiongroup": {
+            "hashes": [
+                "sha256:4d6c0aa6dd825810941c792f53d7b8d71da26f5e5f84f20f9508e8f2d33b140a",
+                "sha256:73866f7f842ede6cb1daa42c4af078e2035e5f7607f0e2c762cc51bb31bbe7b2"
+            ],
+            "markers": "python_version < '3.11'",
+            "version": "==1.0.1"
+        },
         "expects": {
             "hashes": [
                 "sha256:419902ccafe81b7e9559eeb6b7a07ef9d5c5604eddb93000f0642b3b2d594f4c"
@@ -496,6 +504,14 @@
             "markers": "platform_machine == 'x86_64'",
             "version": "==3.5.3"
         },
+        "pkgutil-resolve-name": {
+            "hashes": [
+                "sha256:357d6c9e6a755653cfd78893817c0853af365dd51ec97f3d358a819373bbd174",
+                "sha256:ca27cc078d25c5ad71a9de0a7a330146c4e014c2462d9af19c6b828280649c5e"
+            ],
+            "markers": "python_version < '3.9'",
+            "version": "==1.3.10"
+        },
         "protobuf": {
             "hashes": [
                 "sha256:0f237c1e84e46747397a3e8173edb3116e81163b2878fc944a5193b05876eaee",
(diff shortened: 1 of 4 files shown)


ohare93 (approved):
Looks great :+1:

aplaice:
> Looks great +1

Thanks! (though I'm not too happy about the additional pywin32 module mocking.)

PR #184, opened 2023-01-02
Allow CrowdAnki export for the default 2.1.55+ exporter

### LINKED ISSUE #169: Display CrowdAnki export in the new (non-legacy) export dialog (2.1.55)
With Anki 2.1.54's new export dialog turned on (Tools > Preferences > New import/export handling (beta)), the option to export in CrowdAnki format isn't displayed.

(If this preference isn't turned on, then the option is correctly displayed and export seems to work normally.)

aplaice:
Fix #169.

The old, legacy exporter (Tools > Preferences > Legacy import/export handling) had worked before this commit and now still works.

AFAICT compatibility with Anki 2.1.50+ has been maintained (2.1.50, 2.1.54 and 2.1.55 explicitly tested).

Note that in Anki 2.1.54, CrowdAnki export is only available for the old, then default (non-beta) export method, since the hook that allows easy CrowdAnki to work with the new method was AFAICT only added in 2.1.55.

### DIFF
--- a/crowd_anki/anki/compat/exporting.py
+++ b/crowd_anki/anki/compat/exporting.py
@@ -1,0 +1,58 @@
+"""Compat for Anki 2.1.54-
+
+This is copied verbatim from `qt/aqt/import_export/exporting.py` (in
+Anki 2.1.55).  We need it for our tests, which still assume that we
+have Anki 2.1.26 and where the above module is missing.  We can't
+upgrade our dependency to Anki 2.1.55 which has the module, because
+we're still using Python 3.7 with which latest Anki is incompatible.
+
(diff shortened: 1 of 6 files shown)


aplaice:
I'm not too happy (messy code dealing with multiple Anki versions), but I can't think of any obvious improvements.

(Now also tested with 2.1.56!)

caldotdev:
Anki does not allow me to install the plugin due to an `invalid add-on manifest`

aplaice:
Glad that it's working!

(I also added a `manifest.json` to the new `.ankiaddon` file attached to the release, so hopefully other people won't have issues. (but I added it after your edit, so something else resolved your problem :)).)

PR #195, opened 2023-05-18
Allow testing building crowd_anki/dist when Pipfile.lock is changed

### LINKED ISSUE #182: Add test for running `fetch_dependencies.sh` (building `dist/`)
This should only be called when `Pipfile.lock` is modified, to avoid unnecessary load for GitHub.

This should help us avoiding merging PRs like #176.

<hr/>

Having `dist/` built (and cached) would also be valuable in the future for running end-to-end UI tests (once they're written), so that the CrowdAnki add-on code is exactly in the form it would be on a user's machine.

<hr/>

This should include a test that our dulwich build is pure python.

aplaice:
The uploaded artifact is currently unused.

Fix #182.

We check for changes to both Pipfile and Pipfile.lock to avoid stale a Pipfile.lock.

This will result in some false positive runs of this workflow, when we've, say, added a new script to Pipfile, without changing dependencies.

We will also have false positives when only dev dependencies have been updated.

### DIFF
--- a/.github/workflows/fetch_dependencies.yml
+++ b/.github/workflows/fetch_dependencies.yml
@@ -1,0 +1,38 @@
+name: Fetch dependencies into dist
+
+on:
+  pull_request:
+    paths: [Pipfile, Pipfile.lock]
+  push:
+    branches: [master]
+    paths: [Pipfile, Pipfile.lock]
+
+jobs:
+  fetch_dependencies:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


aplaice:
Ugh, that didn't work (see:

https://github.com/Stvad/CrowdAnki/actions/runs/5040054106/jobs/9038569945?pr=194
)

(Process substitution (`<()`) issues?)

PR #197, opened 2023-06-25
Create action for generating a packaged plugin artifact

aplaice:
This only applies when pushing to a tag.

The yaml files is closely based on the already existing fetch_dependencies workflow.  It doesn't try to re-use the `dist/` artifact generated in the fetch_dependencies workflow to avoid complexity (and because this workflow will be run only rarely, anyway).

* Update generated UI files

* Avoid caching pip wheels when building dist/

Otherwise, if this version of pyyaml was earlier installed (even in a different project!), then pip will use the locally generated and cached pyyaml wheel, despite the `--no-binary`.


aplaice:
For later reference, in case anybody is curious — this doesn't automatically upload the release artifact into the release corresponding to the tag, because AFAIU this requires granting the "write repository" permission to the workflow, which I'd strongly prefer to avoid doing, since that permission also allows modifying the git history (which would be problematic in case of a vulnerability in the GitHub actions/workflow system).

Issue #3, opened 2016-09-01
Require a license statement, and embed it in the JSON


nicolas-raoul:
Having a license lets people fork decks without worrying about legal trouble.
In case of share-alike licenses like CC-BY-SA or GNU-GPL, the license also prevents people from selling a derived deck without publishing it freely too.
So, better require everyone to specify a license, and embed this license in the JSON file.
The license question could be a very simple question when exporting:

```
- Should people who modify this deck make their modifications available to everyone? -> CC-BY-SA-3.0
- Can people do what they want with this deck, including selling a modified version of it? -> MIT
- Nobody is allowed to modify this deck? -> EULA
```
or something similar.

eneller:
add-ons to anki need to be licensed under agpl3 according to the [terms](https://ankiweb.net/account/terms) under "Sharing Add-ons"

PR #202, opened 2023-08-20
Update LICENSE

eneller:
changed the license according to my https://github.com/Stvad/CrowdAnki/issues/3#issuecomment-1685321006 :
add-ons to anki need to be licensed under agpl3 according to the [terms](https://ankiweb.net/account/terms) under "Sharing Add-ons".


Stvad:
Hi thanks for the suggestion! But I don't believe this is required.

to quote specific part of the rules:

> As add-ons extend the computer version, they must be licensed under the AGPL3 or a compatible license.

MIT is compatible with AGPLv3.

PR #213, opened 2024-09-15
Be paranoid about clearing media before copying when exporting

aplaice:
See: 

https://github.com/Stvad/CrowdAnki/pull/212#issuecomment-2351533331

### DIFF
--- a/crowd_anki/export/anki_exporter.py
+++ b/crowd_anki/export/anki_exporter.py
@@ -84,7 +84,7 @@ class AnkiJsonExporter(DeckExporter):
     def _copy_media(self, deck, deck_directory):
         media_directory = deck_directory.joinpath(MEDIA_SUBDIRECTORY_NAME)
 
-        shutil.rmtree(str(media_directory.resolve()), ignore_errors=True)
+        self._clear_media(media_directory.resolve())
         media_directory.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


crowd_anki/export/anki_exporter.py
```diff
@@ -93,3 +93,18 @@ def _copy_media(self, deck, deck_directory):
...
+
+    def _clear_media(self, media_directory: Path):
+        """Clear existing media from export dir before copying.
+
+        Try to be minimally invasive to avoid catastrophic data
+        removal if the relevant variables become incorrectly set
+        elsewhere.
+
+        """
+        unsafe_dirs = [Path.home()]
```
ohare93, line 105: Why end the paranoia there? 😁 include '/' and '' too!

ohare93 (approved):

crowd_anki/export/anki_exporter.py
aplaice: Thanks so much for reviewing/approving! :)

That was the initial version, but had thought that if somebody is using Anki as root then either:

1. they're running in docker, for testing (?), so likely don't have anything too valuable at "/"
2. they like living on the edge.

(Unlike HOME it also doesn't help on Windows, where I have no idea what the appropriate paths would be.)

OTOH it doesn't hurt, so I've added it back in! :)

PR #236, opened 2025-07-06
Prevent card flushing if deck_id does not change

### LINKED ISSUE #228: Prevent update of unchanged notes
## Problem

When importing large decks from JSON, I noticed that Anki will sync every note in the deck despite there being no changes. For very large decks (mine has tens of thousands of notes), this can cause a lot of churn. 

This appears to be caused by `self.anki_object.mod = anki.utils.int_time()` in the note saving process, which unconditionally updates the modification timestamp of all notes.

## Ask

If a note's fields haven't changed from what's already in the collection, don't update the modification timestamp or flush changes to the database.

I'd be happy to push a PR for this feature, if it really is as simple as checking the `self.anki_object` dict for a diff, but I'm not familiar with Anki's backend so I don't know if there would be unintended consequences.

lstrobel:
Fixes #228 by preventing unconditional updates to the collection if the `card.deck_id` does not change on json import.

The default behavior would call `card.flush()` no matter if the card would move or not. I added a boolean output to the  `Card.move_to_deck()` override so that we could detect if there were changes and avoid this behavior.

I also changed the call of `card.flush()` to `collection.update_card()`, as `card.flush()` is marked as deprecated in the current anki library. I checked the implementation, and it seems that `card.flush()` and `collection.update_card()` are functionally identical, so no worries there. 

I tested manually on my local Anki, confirming that card moving behavior is identical to before, and also confirmed that if no cards move, there is no full sync. Given the minimal nature of this change, I feel confident saying that behavior is otherwise identical. Let me know if you would like further testing.

### DIFF
--- a/crowd_anki/anki/overrides/cards.py
+++ b/crowd_anki/anki/overrides/cards.py
@@ -8,15 +8,24 @@ def move_to_deck(self, deck_id, move_from_dynamic_deck=False):
     :parameter move_from_dynamic_deck Specifies if we should perform move if card is in dynamic deck
      or should we just change it's "odid" old/original deck id.
      The function is not doing flush, so you need to call flush yourself in order to persist changes into DB
+    :return: True if card was changed, False otherwise.
     """
+    changed = False
     if move_from_dynamic_deck:
         self.col.sched.remFromDyn([self.id])
         self.load()
+        changed = True
 
     if self.odid:
-        self.odid = deck_id
+        if self.odid != deck_id:
+            self.odid = deck_id
+            changed = True
     else:
-        self.did = deck_id
+        if self.did != deck_id:
+            self.did = deck_id
+            changed = True
+
+    return changed
 
 
 Card.move_to_deck = move_to_deck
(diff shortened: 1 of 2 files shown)


lstrobel:
Your understanding is correct, and to be honest I only made that change because I have no idea what it does and figured better safe (flush) than sorry - `move_to_deck()` is only called via the note anyways.

Ill push a change to address this

lstrobel:
I ended up just removing the parameter. Let me know if you're alright with that approach.

aplaice:
Yes, I agree, this is probably the cleanest approach!

(I _think_ that moving `if move_from_dynamic_deck: self.col.sched.remFromDyn ...` directly into the `if self.odid` branch should do the job (if `odid` is defined then we're in a dynamic deck), but I'm not 100% sure and I don't think it's worth having stale, unused code.)

aplaice (approved):
Manually testing doesn't reveal any issues. (Updating notes and moving cards back into the original deck both work. I haven't noticed any new problems with sync.)

TBH I haven't noticed any speed-up in AnkiWeb sync after re-import, but all my CrowdAnki decks are sufficiently small that sync after re-import is already near-instantaneous.


<hr/>

Thanks again! Sorry for the delay!

PR #255, opened 2026-06-02
Fix import dialog showing 0 notes/media for subdeck-only decks

### LINKED ISSUE #190: Note count in import dialog doesn't include subdecks
(Low priority.)

### To reproduce:

1. Import the following minimal deck with one subdeck and one card in the subdeck (and no cards in the main deck):

<details>

```json
{
    "__type__": "Deck",
    "children": [
        {
            "__type__": "Deck",
            "children": [],
            "crowdanki_uuid": "12684a9e-b15b-11ed-985f-c9e14aaf9362",
            "deck_config_uuid": "12684f30-b15b-11ed-985f-c9e14aaf9362",
            "desc": "",
            "dyn": 0,
            "extendNew": 0,
            "extendRev": 0,
            "media_files": [],
            "name": "subdeck",
            "newLimit": null,
            "newLimitToday": null,
            "notes": [
                {
                    "__type__": "Note",
                    "fields": [
                        "a",
                        "b"
                    ],
                    "guid": "XRzqD(ea!",
                    "note_model_uuid": "22b980ca-b15b-11ed-b503-c5b4cccbc2bf",
                    "tags": []
                }
            ],
            "reviewLimit": null,
            "reviewLimitToday": null
        }
    ],
    "crowdanki_uuid": "1268232a-b15b-11ed-985f-c9e14aaf9362",
    "deck_config_uuid": "12684f30-b15b-11ed-985f-c9e14aaf9362",
    "deck_configurations": [
        {
            "__type__": "DeckConfig",
            "autoplay": true,
            "buryInterdayLearning": false,
            "crowdanki_uuid": "12684f30-b15b-11ed-985f-c9e14aaf9362",
            "dyn": false,
            "interdayLearningMix": 0,
            "lapse": {
                "delays": [
                    10.0
                ],
                "leechAction": 1,
                "leechFails": 8,
                "minInt": 1,
                "mult": 0.0
            },
            "maxTaken": 60,
            "name": "Default",
            "new": {
                "bury": false,
                "delays": [
                    1.0,
                    10.0
                ],
                "initialFactor": 2500,
                "ints": [
                    1,
                    4,
                    0
                ],
                "order": 1,
                "perDay": 20
            },
            "newGatherPriority": 0,
            "newMix": 0,
            "newPerDayMinimum": 0,
            "newSortOrder": 0,
            "replayq": true,
            "rev": {
                "bury": false,
                "ease4": 1.3,
                "hardFactor": 1.2,
                "ivlFct": 1.0,
                "maxIvl": 36500,
                "perDay": 200
            },
            "reviewOrder": 0,
            "timer": 0
        }
    ],
    "desc": "",
    "dyn": 0,
    "extendNew": 0,
    "extendRev": 0,
    "media_files": [],
    "name": "Default",
    "newLimit": null,
    "newLimitToday": null,
    "note_models": [
        {
            "__type__": "NoteModel",
            "crowdanki_uuid": "22b980ca-b15b-11ed-b503-c5b4cccbc2bf",
            "css": ".card {\n    font-family: arial;\n    font-size: 20px;\n    text-align: center;\n    color: black;\n    background-color: white;\n}\n",
            "flds": [
                {
                    "collapsed": false,
                    "description": "",
                    "font": "Arial",
                    "name": "Front",
                    "ord": 0,
                    "plainText": false,
                    "rtl": false,
                    "size": 20,
                    "sticky": false
                },
                {
                    "collapsed": false,
                    "description": "",
                    "font": "Arial",
                    "name": "Back",
                    "ord": 1,
                    "plainText": false,
                    "rtl": false,
                    "size": 20,
                    "sticky": false
                }
            ],
            "latexPost": "\\end{document}",
            "latexPre": "\\documentclass[12pt]{article}\n\\special{papersize=3in,5in}\n\\usepackage[utf8]{inputenc}\n\\usepackage{amssymb,amsmath}\n\\pagestyle{empty}\n\\setlength{\\parindent}{0in}\n\\begin{document}\n",
            "latexsvg": false,
            "name": "Basic",
            "req": [
                [
                    0,
                    "any",
                    [
                        0
                    ]
                ]
            ],
            "sortf": 0,
            "tmpls": [
                {
                    "afmt": "{{FrontSide}}\n\n<hr id=answer>\n\n{{Back}}",
                    "bafmt": "",
                    "bfont": "",
                    "bqfmt": "",
                    "bsize": 0,
                    "did": null,
                    "name": "Card 1",
                    "ord": 0,
                    "qfmt": "{{Front}}"
                }
            ],
            "type": 0
        }
    ],
    "notes": [],
    "reviewLimit": null,
    "reviewLimitToday": null
}
```
</details>

2. Read the import dialog.

### Expected behaviour

The "Deck Parts to Use" says:

```
Notes: 1.
```

### Actual behaviour

The "Deck Parts to Use" says:

```
Notes: 0.
```

<hr/>

This is because we currently only check in [the main deck itself](https://github.com/Stvad/CrowdAnki/blob/master/crowd_anki/importer/import_dialog.py#L141).

If we had a CrowdAnki deck object (rather than a pure JSON), we could use the `deck.get_note_count` function, so perhaps we should carry out the ["deck initialization from JSON" (i.e. the conversion from JSON to "internal CrowdAnki object")](https://github.com/Stvad/CrowdAnki/blob/master/crowd_anki/importer/anki_importer.py#L38) before showing the dialog?

This conversion itself doesn't touch the user's collection, we don't use any information from the import dialog in this conversion and there's nothing in the final CrowdAnki deck object that's not in the initial JSON. Also, the earlier we do the conversion, the less dependent we're on the specifics of the JSON implementation if we end up having alternative storage formats.

The main con might be that the conversion might be slow? (I don't think it is — AFAIU most of our slowness comes from interfacing with the Anki database — i.e. converting from our internal object to Anki's.)

(I believe the same applies to media counts.)

fabiomanz:
## Summary

Fixes #190. The import dialog's "Notes:" and "Media Files:" counts were read directly from the top-level JSON arrays, so decks whose notes/media live only in subdecks displayed `0`.

This matches the approach proposed in the issue: do the JSON→`Deck` conversion once, *before* showing the dialog, and let the dialog ask the `Deck` for the counts. This removes a coupling between the dialog and the on-disk JSON layout.

## Changes

- `Deck`: new `get_media_file_count()` parallel to the existing `get_note_count()`, plus a `media_files` field.
- `deck_initializer.from_json`: populate `deck.media_files` from the JSON before `post_import_filter` strips the key.
- `anki_importer.load_deck`: build the `Deck` up front and pass it to the dialog instead of the raw JSON.
- `ImportDialog`: takes a `Deck`; note models are read from `deck.metadata.models`; counts come from the two `get_*_count()` methods.

## Tests

- `test/representation/deck_spec.py` (new) — unit tests for both count methods (flat, nested, subdeck-only).
- `test/representation/deck_initializer_spec.py` — added an integration test verifying `from_json` populates `media_files` recursively.

## Manual verification

Tested locally in Anki by importing a deck whose notes live only in a subdeck — the dialog now reports the correct counts instead of `0`.

### DIFF
--- a/crowd_anki/importer/anki_importer.py
+++ b/crowd_anki/importer/anki_importer.py
@@ -27,15 +27,15 @@ class AnkiJsonImporter:
         :param directory_path: Path
         """
         deck_json = self.read_deck(self.get_deck_path(directory_path))
+        deck = deck_initializer.from_json(deck_json)
 
-        import_config = self.read_import_config(directory_path, deck_json)
+        import_config = self.read_import_config(directory_path, deck)
         if import_config is None:
             return False
 
         if aqt.mw:
             aqt.mw.create_backup_now()
         try:
-            deck = deck_initializer.from_json(deck_json)
             deck.save_to_collection(self.collection, import_config=import_config)
 
             if import_config.use_media:
@@ -84,7 +84,7 @@ class AnkiJsonImporter:
             return json.load(deck_file)
 
     @staticmethod
-    def read_import_config(directory_path, deck_json):
+    def read_import_config(directory_path, deck):
         file_path = directory_path.joinpath(IMPORT_CONFIG_NAME)
 
         if not file_path.exists():
@@ -93,7 +93,7 @@ class AnkiJsonImporter:
             with file_path.open(encoding='utf8') as meta_file:
                 import_dict = yaml.full_load(meta_file)
 
-        import_dialog = ImportDialog(deck_json, import_dict)
+        import_dialog = ImportDialog(deck, import_dict)
         if import_dialog.exec() == QDialog.DialogCode.Rejected:
             return None
 
--- a/crowd_anki/importer/import_dialog.py
+++ b/crowd_anki/importer/import_dialog.py
@@ -66,13 +66,13 @@ class ImportConfig(PersonalFieldsHolder):
 
 
 class ImportDialog(QDialog):
-    def __init__(self, deck_json, config, parent=None):
+    def __init__(self, deck, config, parent=None):
         super().__init__(None)
         self.parent = parent
         self.form = ConfigUI()
         self.form.setupUi(self)
         self.userConfig = ConfigSettings.get_instance()
-        self.deck_json = deck_json
+        self.deck = deck
         self.import_defaults = ImportDefaults.from_dict(config)
         self.personal_field_ui_dict = defaultdict(dict)
         self.ui_initial_setup()
@@ -120,12 +120,12 @@ class ImportDialog(QDialog):
                     item.setCheckState(Qt.CheckState.Checked)
         self.form.list_personal_fields.itemClicked.connect(on_click)
 
-        for model in self.deck_json["note_models"]:
-            model_name = model["name"]
-            model_id = model[UUID_FIELD_NAME]
+        for model in self.deck.metadata.models.values():
+            model_name = model.anki_dict["name"]
+            model_id = model.anki_dict[UUID_FIELD_NAME]
             add_header(model_name)
 
-            for field in model["flds"]:
+            for field in model.anki_dict["flds"]:
                 field_name = field["name"]
                 field_ui = add_field(field_name, self.import_defaults.is_personal_field(model_name, field_name))
                 self.personal_field_ui_dict[model_name].setdefault(field_name, field_ui)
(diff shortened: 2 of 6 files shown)


aplaice:
Thanks very much for this! (And thanks for patiently rebasing over master, multiple times...)

I'm sorry for the horrible delay replying and also for the multiple issues with our actions!

One slight issue is that it seems that if the same media files are present in multiple sub-decks (e.g. re-using the same media in multiple notes or having [shared (`_xxx`) media](https://docs.ankiweb.net/media.html#checking-media)) then I think we'll get an overcount.  (Manual testing seems to confirm this.)  (FWIW I think that using `'file'` as the universal media filename in `_make_deck`, in the `deck_spec` test is an oversimplification, since if the filename is the same we should only really report a single media file.)

Otherwise, this looks great!

fabiomanz:
Thanks for the feedback. Can you please take a look if the newest changes fulfill your questions?

aplaice:
Thanks very much!

